PR #4, opened 2020-07-11
Add demo gif

JoannaLXY:


innerlee:
still large.

may shorten the first shot and the second shot a little bit. and also decrease the fps to 10 or 8

PR #22, opened 2020-07-22
setup github action

jin-s13:
Init github action.


mmpose/models/backbones/resnet.py
```diff
@@ -85,13 +85,16 @@ def __init__(self,
 
     @property
     def norm1(self):
+        """nn.Module: the normalization layer named "norm1" """
```
JoannaLXY, line 88: The meaning of this is not clear enough. Does it mean "Get the name of normalization layer "norm1" "?

mmpose/models/keypoint_heads/top_down_multi_stage_head.py
```diff
@@ -82,7 +82,10 @@ def __init__(self,
             self.multi_final_layers.append(final_layer)
 
     def forward(self, x):
-        # return a list of heatmaps from multiple stages.
+        """Forward function.
+
+        return a list of heatmaps from multiple stages.
```
JoannaLXY, line 87: Use the correct format for return.

mmpose/models/detectors/base.py
```diff
@@ -84,6 +88,31 @@ def _parse_losses(losses):
...
+        hook. Note that in some complicated cases or models, the whole process
+        including back propagation and optimizer updating is also defined in
+        this method, such as GAN.
+
+        Args:
+            data_batch (dict): The output of dataloader.
+            optimizer (:obj:`torch.optim.Optimizer` | dict): The optimizer of
+                runner is passed to ``train_step()``. This argument is unused
+                and reserved.
+
```
JoannaLXY, line 104: Add `kwargs`

PR #26, opened 2020-07-23
add multiple backbones from mmclassification

jin-s13:


innerlee:
Will the added configs own its checkpoint in modelzoo?

jin-s13:
> Will the added configs own its checkpoint in modelzoo?

Yes. We will add the checkpoint and setup the README.md for each model soon.

innerlee:
cool

PR #27, opened 2020-07-23
update demos

jin-s13:


demo/img_demo.py
```diff
@@ -74,6 +74,8 @@ def main():
         if args.out_img_root == '':
             out_file = None
         else:
+            if not os.path.isdir(args.out_img_root):
```
innerlee, line 77: check this https://docs.python.org/3/library/os.html#os.makedirs

innerlee: how about setting `exist_ok=True`

jin-s13: OK.

PR #33, opened 2020-07-24
unify use_gt in configs

jin-s13:


innerlee:
is this a typo-fix?

jin-s13:
Some cfgs use_gt=False, while others use_gt=True.  Better use the same setting.

innerlee:
Does this setting affect training? Will they also affect provided weights in modelzoo?

jin-s13:
Just the evaluation settings. This will not affect training and the provided weights. 
However, in the accompanied log, the testing accuracy will be different...

jin-s13:
No. Only the log will be affected. Model zoo will not.

PR #34, opened 2020-07-24
device-agnostic code

jin-s13:


mmpose/models/detectors/bottom_up.py
```diff
@@ -194,7 +194,6 @@ def forward_test(self, img, img_metas, **kwargs):
         tags_list = []
         for idx, s in enumerate(sorted(test_scale_factor, reverse=True)):
             image_resized = aug_data[idx]
-            image_resized = image_resized.cuda(non_blocking=True)
```
innerlee, line 197: is `image_resized` on gpu already?

tests/test_model/test_top_down_forward.py
```diff
@@ -43,10 +43,11 @@ def test_topdown_forward():
     assert isinstance(losses, dict)
 
     # Test forward test
-    with torch.no_grad():
-        detector = detector.cuda()
-        _ = detector.forward(
-            imgs.cuda(), img_metas=img_metas, return_loss=False)
+    if torch.cuda.is_available():
```
innerlee, line 46: else?

innerlee: Tests that require cuda could be annotated with some kind of python decorators.

innerlee: btw why train does not need .cuda()

mmpose/models/losses/multi_loss_factory.py
```diff
@@ -9,20 +9,19 @@
...
 
     Args:
         t (torch.Tensor): input
         requires_grad (bool): Option to use requires_grad.
-        need_cuda (bool): Opthin to use cuda.
+        device: torch device
     Returns:
         inp (torch.Tensor): zero input.
     """
     inp = torch.autograd.Variable(t, requires_grad=requires_grad)
```
innerlee, line 22: Unrelated to this pr, but Variable has been deprecated. https://pytorch.org/docs/stable/autograd.html#variable-deprecated

innerlee: Maybe we can remove all usage of Variables in a separate PR.

jin-s13: Good idea.

PR #36, opened 2020-07-28
Update modelzoo

jin-s13:
add some checkpoints and logs.


configs/top_down/alexnet/README.md
```diff
@@ -0,0 +1,20 @@
+# Imagenet classification with deep convolutional neural networks
```
innerlee, line 1: does top-down methods only differs by backbone?

jin-s13: Yes. For most top-down methods, the network architectures are the main differences.

innerlee: 🤔

innerlee: let's merge it as is, and change only when necessary

innerlee:
Could you add the newly added models to the repo's readme?

jin-s13:
> Could you add the newly added models to the repo's readme?

Sure.

PR #38, opened 2020-07-28
Update modelzoo

jin-s13:


innerlee:
may need to rebase master

innerlee:
The diff looks very strange: maybe something went wrong when doing rebase. Let me try

PR #45, opened 2020-07-31
Update runtime.txt

innerlee:


innerlee:
@jin-s13 could you verify that we can run using the latest pillow? v7.1.0 i guess

jin-s13:
We use pillow==6.2.2 to be compatible with [mmdetection requirement](https://github.com/open-mmlab/mmdetection/blob/master/requirements/runtime.txt), which requires `Pillow<=6.2.2`

innerlee:
@hellock can mmdet upgrade pillow to higher version?

hellock:
> @hellock can mmdet upgrade pillow to higher version?

This version requirement can be removed.

PR #50, opened 2020-08-04
Update README.md

jin-s13:
fix typos.


innerlee:
the diff shows typo fixes in readme, and nothing in test/data has changed.

jin-s13:
> the diff shows typo fixes in readme, and nothing in test/data has changed.

Yes. Only the typos.

innerlee:
not related to this pr, but how about adding links to their corresponding configs on the names, like

image

jin-s13:
> not related to this pr, but how about adding links to their corresponding configs on the names, like
> 
> image

Let me try.

innerlee:
For the links, write the path from the repo root, which starts with a leading slash /. in this way, the docs site would be smart enough to link back to this repo.

image

PR #63, opened 2020-08-13
fix bugs and typos

jin-s13:
1. `chmod +x` to the shell files. 
2. fix typos 
3. fix bugs in bottom-up post-processing


mmpose/core/post_processing/group.py
```diff
@@ -249,33 +249,33 @@ def adjust(self, ans, heatmaps):
         """
         N, K, H, W = heatmaps.shape
         for batch_id, people in enumerate(ans):
```
innerlee, line 251: it seems that loops are error prone. can this be rewritten by tensor operation?

innerlee: not easy

PR #64, opened 2020-08-13
add model zoo

jin-s13:
update readme.md and add model zoo


innerlee:
nice, so the mpii dataset pr is not blocked by accuracy check any more, right?

jin-s13:
YES :)

PR #70, opened 2020-08-16
OChuman dataset

jin-s13:


mmpose/datasets/datasets/top_down/topdown_ochuman_dataset.py
```diff
@@ -0,0 +1,282 @@
...
+        self.classes = ['__background__'] + cats
+        self.num_classes = len(self.classes)
+        self._class_to_ind = dict(zip(self.classes, range(self.num_classes)))
+        self._class_to_coco_ind = dict(zip(cats, self.coco.getCatIds()))
+        self._coco_ind_to_class_ind = dict([(self._class_to_coco_ind[cls],
+                                             self._class_to_ind[cls])
+                                            for cls in self.classes[1:]])
+        self.image_set_index = self.coco.getImgIds()
+        self.num_images = len(self.image_set_index)
+        self.id2name, self.name2id = _get_mapping_id_name(self.coco.imgs)
```
innerlee, line 121: take this as a game, but this function is a one-liner
`id2name = {id:image['file_name'] for id, image in imgs.items()}`

mmpose/datasets/datasets/top_down/topdown_ochuman_dataset.py
```diff
@@ -0,0 +1,282 @@
...
+
+        # sanitize bboxes
+        valid_objs = []
+        for obj in objs:
+            x, y, w, h = obj['bbox']
+            x1 = max(0, x)
+            y1 = max(0, y)
+            x2 = min(width - 1, x1 + max(0, w - 1))
+            y2 = min(height - 1, y1 + max(0, h - 1))
+            if obj['area'] > 0 and x2 >= x1 and y2 >= y1:
```
innerlee, line 160: does the new area need to be updated and checked?

jin-s13: Here we only ignore some invalid training data; no need to update the `area'.
`area' is the area of the object mask, we do not need it in training and should not update it either.

innerlee: okay

innerlee: but do we need to check that `(x2-x1)*(y2-y1)!=0`?

mmpose/datasets/datasets/top_down/topdown_ochuman_dataset.py
```diff
@@ -0,0 +1,282 @@
...
+                valid_objs.append(obj)
+        objs = valid_objs
+
+        rec = []
+        for obj in objs:
+            if max(obj['keypoints']) == 0:
+                continue
+            joints_3d = np.zeros((num_joints, 3), dtype=np.float)
+            joints_3d_visible = np.zeros((num_joints, 3), dtype=np.float)
+            for ipt in range(num_joints):
```
innerlee, line 171: is this for loop equivalent to
```python
keypoints = anno['keypoints'].reshape(-1, 3)
joints_3d[:, :2] = keypoints[:, :2]
joints_3d_visible[:, :2] = np.minimum(1, keypoints[:, 2:3])
```

innerlee: note: need to change the first line if anno['keypoints'] is not 3.

mmpose/datasets/datasets/top_down/topdown_ochuman_dataset.py
```diff
@@ -0,0 +1,282 @@
...
+                if t_vis > 1:
+                    t_vis = 1
+                joints_3d_visible[ipt, 0] = t_vis
+                joints_3d_visible[ipt, 1] = t_vis
+                joints_3d_visible[ipt, 2] = 0
+
+            center, scale = self._xywh2cs(*obj['clean_bbox'][:4])
+            rec.append({
+                'image_file':
+                os.path.join(self.img_prefix, self.id2name[index]),
```
innerlee, line 185: if we put this long line `os.path.join(self.img_prefix, self.id2name[index])` before append, then this append block might have a prettier format

mmpose/datasets/datasets/top_down/topdown_ochuman_dataset.py
jin-s13: Nice~

mmpose/datasets/datasets/top_down/topdown_ochuman_dataset.py
jin-s13: Guess `obj['area'] > 0` already meets `(x2-x1)*(y2-y1)!=0`?

Issue #71, opened 2020-08-17
KeyError: 'rotation'


CUIcaicai:
where can i download this checkpoint。

and i use hrnet_w48_coco_256x192-b9e0b3ab_20200708.pth
but i get the error：
  File "demo/top_down_video_demo_with_mmdet.py", line 118, in <module>
    main()
  File "demo/top_down_video_demo_with_mmdet.py", line 91, in main
    format='xyxy')
  File "/home/cl/gitclone/mmpose/mmpose/apis/inference.py", line 228, in inference_pose_model
    pose = _inference_single_pose_model(model, img_or_path, bbox)
  File "/home/cl/gitclone/mmpose/mmpose/apis/inference.py", line 175, in _inference_single_pose_model
    data = test_pipeline(data)
  File "/home/cl/gitclone/mmpose/mmpose/datasets/pipelines/shared_transform.py", line 78, in __call__
    data = t(data)
  File "/home/cl/gitclone/mmpose/mmpose/datasets/pipelines/top_down_transform.py", line 181, in __call__
    r = results['rotation']
KeyError: 'rotation'

innerlee:
Could try the latest master?

innerlee:
sorry, the fix was in this pr https://github.com/open-mmlab/mmpose/pull/69
i will take a look

jin-s13:
Sorry for the bugs.
Please take a look at [this commit](https://github.com/open-mmlab/mmpose/pull/69/commits/1c3ae8e35016cc11c96bd17b92adc51a5b98874b)

CUIcaicai:
Ok，thanks，i will try it




On 08/17/2020 19:31, Jas wrote:

Sorry for the bugs.
Please take a look at this commit

Issue #29, opened 2020-07-23
Support Demoing HigherHRNet


ThorJonsson:
Hi, 

Thank you for your interesting work. I am trying to test HigherHRNet on a collection of videos I've found very challenging for existing human pose estimation models. However when I try to execute the following command:

`python demo/video_demo_with_mmdet.py $SCRATCH/mmdetection/configs/faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py $SCRATCH/mmdetection/checkpoints/faster_rcnn_r50_fpn_1x_coco_20200130-047c8118.pth   configs/bottom_up/higherhrnet/coco/higher_hrnet48_coco_512x512.py   checkpoints/higher_hrnet48_coco_512x512-60fedcbc_20200712.pth    --video-path $SCRATCH/VIBE/sample_video.mp4     --show     --out-video-root ./`

I get a number of config related issues, the first is that the image size seems to be treated as both an integer as well as a list, I made a temporary fix with a try except statement, but the second issue I'm dealing with now is the lack of a flip_index.

I am not sure what purpose the flip_index has but I've tried making it 0 or [0,1] as I saw in another example and neither of these work. 

For a detailed traceback I get

```python
Traceback (most recent call last):
  File "demo/video_demo_with_mmdet.py", line 117, in <module>
    main()
  File "demo/video_demo_with_mmdet.py", line 90, in main
    format='xyxy')
  File "/mmpose/mmpose/apis/inference.py", line 238, in inference_pose_model
    pose = _inference_single_pose_model(model, img_or_path, bbox)
  File "mmpose/mmpose/apis/inference.py", line 194, in _inference_single_pose_model
    return_loss=False, img=data['img'], img_metas=data['img_metas'])
  File ".conda/envs/pytorch150/lib/python3.7/site-packages/torch/nn/modules/module.py", line 550, in __call__
    result = self.forward(*input, **kwargs)
  File "mmpose/mmpose/models/detectors/bottom_up.py", line 103, in forward
    return self.forward_test(img, img_metas, **kwargs)
  File "mmpose/mmpose/models/detectors/bottom_up.py", line 218, in forward_test
    self.test_cfg['flip_test'])
  File "mmpose/mmpose/core/evaluation/bottom_up_eval.py", line 157, in aggregate_results
    heatmaps[1]) / 2.0 if flip_test else heatmaps[0]
```

Hope you can help me, I'm really excited about your work and hoping to be able to make contributions in the future.

[...]

PR #69, opened 2020-08-16
support bottom-up demo

jin-s13:
#29

PR #72, opened 2020-08-17
support bottom-up demo

innerlee:
Fix https://github.com/open-mmlab/mmpose/issues/29 and fix https://github.com/open-mmlab/mmpose/issues/71

### DIFF
--- a/mmpose/models/detectors/bottom_up.py
+++ b/mmpose/models/detectors/bottom_up.py
@@ -267,7 +267,9 @@ class BottomUp(BasePose):
             kpt_score_thr (float, optional): Minimum score of keypoints
                 to be shown. Default: 0.3.
             pose_kpt_color (np.array[Nx3]`): Color of N keypoints.
+                If None, do not draw keypoints.
             pose_limb_color (np.array[Mx3]): Color of M limbs.
+                If None, do not draw limbs.
             thickness (int): Thickness of lines.
             font_scale (float): Font scales of texts.
             win_name (str): The window name.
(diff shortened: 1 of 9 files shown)


innerlee:
I cherry-picked the demo codes from https://github.com/open-mmlab/mmpose/pull/69, to make the diff more readable.
@jin-s13 pls double check

mmpose/models/detectors/bottom_up.py
```diff
@@ -236,3 +243,117 @@ def forward_test(self, img, img_metas, **kwargs):
...
+        for res in result:
+            pose_result.append(res['keypoints'])
+
+        for person_id, kpts in enumerate(pose_result):
+            # draw each point on image
+            for kid, kpt in enumerate(kpts):
+                x_coord, y_coord, kpt_score = int(kpt[0]), int(kpt[1]), kpt[2]
+                if kpt_score > kpt_score_thr:
+                    # cv2.circle(img, (x_coord, y_coord), radius,
+                    #            pose_kpt_color, thickness)
```
innerlee, line 297: so `pose_kpt_color` is not used?

jin-s13: Shall we uncomment it?

mmpose/models/detectors/bottom_up.py
```diff
@@ -236,3 +243,117 @@ def forward_test(self, img, img_metas, **kwargs):
         image_path.extend(img_metas['image_file'])
 
         return results, scores, image_path
+
+    def show_result(self,
+                    img,
+                    result,
+                    skeleton=None,
+                    kpt_score_thr=0.3,
+                    pose_kpt_color=None,
```
innerlee, line 252: since `None` is not a valid value for `pose_kpt_color` and `pose_limb_color` in this function, it is better to either
* do not provide defaults in the signature
* assert that they are not None
* provide default colors when None

innerlee: I prefer option 3 btw

jin-s13: change to "draw kpt&limbs, only if the color is not None".
And add `assert` to the color.

PR #85, opened 2020-08-25
metrics -> metric

liuxin9608:
metrics -> metric

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -21,6 +21,7 @@ def parse_args():
     parser.add_argument(
         '--eval',
         default='mAP',
+        nargs='+',
         help='evaluation metric, which depends on the dataset,'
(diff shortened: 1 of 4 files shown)


jin-s13 (approved):

tools/test.py
```diff
@@ -21,6 +21,7 @@ def parse_args():
     parser.add_argument(
         '--eval',
         default='mAP',
+        nargs='+',
```
innerlee, line 24: why multiple args?

liuxin9608: In hand pose estimation, we will need multiple metrics, such as PCK, AUC...
We can choose to use different metrics.

innerlee: then why change metric to metrics?
doesn't that mean we want the user to specify one metric?

innerlee (approved):

PR #121, opened 2020-09-03
consistent between docs and demo code

zhouhang95:


jin-s13:
Thanks!
Maybe `out-video-root` is a better name.

docs/getting_started.md
```diff
@@ -176,7 +176,7 @@ python demo/top_down_video_demo_with_mmdet.py mmdetection/configs/faster_rcnn/fa
     mmdetection/checkpoints/faster_rcnn_r50_fpn_1x_coco_20200130-047c8118.pth \
     configs/top_down/hrnet/coco/hrnet_w48_coco_256x192.py \
     hrnet_w48_coco_256x192-b9e0b3ab_20200708.pth \
-    --video_path demo/demo_video.mp4 \
+    --video-path demo/demo_video.mp4 \
     --output-video-root vis_results
```
jin-s13, line 180: Better to change this `output-video-root` to `out-video-root`

demo/bottom_up_video_demo.py
```diff
@@ -19,7 +19,7 @@ def main():
         default=False,
         help='whether to show visualizations.')
     parser.add_argument(
-        '--out-video-root',
+        '--output-video-root',
```
jin-s13, line 22: It is better to keep this name as it is, and change the docs.

jin-s13 (approved):

innerlee:
We can make the scripts more consistent by imposing this `abc-def-gh`  convention on all scripts in tools folder. like here https://github.com/open-mmlab/mmpose/blob/master/tools/test.py#L28

it can be done in a separate pr

Issue #123, opened 2020-09-07
MPII dataset processing


syusukee:
It's a really fantastic work!  Can you share the processing code on how to transform the .mat file to json file? I've found the annotations is in the type of .mat file on the MPII Dataset website. Thanks a lot!!

jin-s13:
Thanks for your appreciation. We will add this tool soon.

PR #126, opened 2020-09-09
add mat2json tools

jin-s13:
#123


innerlee:
could you add some examples in doc?

innerlee:
ideally also in test

innerlee (approved):

PR #127, opened 2020-09-11
Update key indicator

jin-s13:
Fix key errors & update key indicator (support lower case now.)

### DIFF
--- a/mmpose/core/evaluation/eval_hooks.py
+++ b/mmpose/core/evaluation/eval_hooks.py
@@ -32,7 +32,7 @@ class EvalHook(Hook):
 
     rule_map = {'greater': lambda x, y: x > y, 'less': lambda x, y: x < y}
     init_value_map = {'greater': -inf, 'less': inf}
-    greater_keys = ['acc', 'AP', 'PCK', 'AUC']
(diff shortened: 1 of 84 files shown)


mmpose/core/evaluation/eval_hooks.py
```diff
@@ -32,15 +32,15 @@ class EvalHook(Hook):
 
     rule_map = {'greater': lambda x, y: x > y, 'less': lambda x, y: x < y}
     init_value_map = {'greater': -inf, 'less': inf}
-    greater_keys = ['acc', 'mAP', 'PCK', 'AUC']
-    less_keys = ['loss', 'EPE']
+    greater_keys = ['acc', 'ap', 'ar', 'pck', 'auc']
```
innerlee, line 35: why not add 'map' in this list

jin-s13: If a key includes 'ap' (e.g., AP, AP 0.5, AP small), then it belongs to the group of greater_keys.

innerlee: i mean, can we avoid changing the 80 many files?

innerlee (approved):

Issue #138, opened 2020-09-19
Slow inference of bottom-up models


Witek-:
I am getting very slow performance on a single person video on RTX2080Ti with the following command

python demo/bottom_up_video_demo.py     configs/bottom_up/resnet/coco/res152_coco_512x512.py res152_coco_512x512-364eb38d_20200822.pth     --video-path demo/tabata1_640x360.mp4 --show --device cuda:0

FPS ~2.5. It was calculated as total processing time divided by the number of video frames, so it is not super precise...

Any idea what might be wrong?

innerlee:
Inference speedup is on our list https://github.com/open-mmlab/mmpose/issues/9#issuecomment-662836766
To gather more info, could you tell what's the expected speed in this case?

jin-s13:
For higher inference speed, you may try to set `adjust=False`, `refine=False`, `flip_test=False` in the config file.

innerlee:
maybe we should add a section in doc to explain how to set flags to enable fast demo, and how to set them to get accurate demo.

Witek-:
> For higher inference speed, you may try to set `adjust=False`, `refine=False`, `flip_test=False` in the config file.

I divided the number of frames by the entire 'while' loop time, so fps is calculated more accurately.

standard config - 2.4 fps
adjust=False - 2.5 fps - marginal gain
refine=False - 10.3 fps
adjust=False and refine=False - 10.4 fps


However, when I use flip_test=False get the following error:

Traceback (most recent call last):
  File "demo/bottom_up_video_demo.py", line 104, in <module>
    main()
  File "demo/bottom_up_video_demo.py", line 71, in main
    pose_results = inference_bottom_up_pose_model(pose_model, img)
  File "/home/witek/Desktop/open-mmlab/mmpose/mmpose/apis/inference.py", line 322, in inference_bottom_up_pose_model
    return_loss=False, img=data['img'], img_metas=data['img_metas'])
  File "/home/witek/miniconda3/envs/open-mmlab/lib/python3.7/site-packages/torch/nn/modules/module.py", line 722, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/home/witek/Desktop/open-mmlab/mmpose/mmpose/models/detectors/bottom_up.py", line 112, in forward
    return self.forward_test(img, img_metas, **kwargs)
  File "/home/witek/Desktop/open-mmlab/mmpose/mmpose/models/detectors/bottom_up.py", line 241, in forward_test
[...]

PR #142, opened 2020-09-21
fix assertion

jin-s13:
#138 fix assertion error.


mmpose/core/post_processing/post_transforms.py
```diff
@@ -105,7 +106,7 @@ def transform_preds(coords, center, scale, output_size):
     Returns:
         np.ndarray: Predicted coordinates in the images.
     """
-    assert coords.shape[1] == 2 or coords.shape[1] == 5
+    assert coords.shape[1] == 2 or coords.shape[1] == 4 or coords.shape[1] == 5
```
innerlee, line 109: `coords.shape[1] in (2, 4, 5)`

innerlee (approved):

PR #157, opened 2020-09-25
Add the function of Procrutes Alignment

zengwang430521:
Add the function of Procrutes Alignment used in 3D joints error evaluation and corresponding test codes.

### DIFF
--- a/mmpose/core/evaluation/mesh_eval.py
+++ b/mmpose/core/evaluation/mesh_eval.py
@@ -1,0 +1,68 @@
+# ------------------------------------------------------------------------------
+# Adapted from https://github.com/nkolot/SPIN
+# Original licence: Copyright (c) 2018 akanazawa, under the MIT License.
+# ------------------------------------------------------------------------------
+
+import numpy as np
+
+
+def compute_similarity_transform(source_points, target_points):
+    """Computes a similarity transform (sR, t) that takes a set of 3D points
+    source_points (N x 3) closest to a set of 3D points target_points, where R
+    is an 3x3 rotation matrix, t 3x1 translation, s scale. And return the
+    transformed 3D points source_points_hat (N x 3). i.e. solves the orthogonal
+    Procrutes problem.
+
+    Notes:
+        Points number: N
+
+    Args:
+        source_points (np.ndarray([N, 3])): Source point set.
+        target_points (np.ndarray([N, 3])): Target point set.
+
+    Returns:
+        source_points_hat (np.ndarray([N, 3])): Transformed source point set.
+    """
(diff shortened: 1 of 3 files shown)


mmpose/core/evaluation/mesh_eval.py
```diff
@@ -0,0 +1,68 @@
...
+
+    transposed = False
+    if source_points.shape[0] != 3 and source_points.shape[0] != 2:
+        source_points = source_points.T
+        target_points = target_points.T
+        transposed = True
+    assert (target_points.shape[1] == source_points.shape[1])
+
+    # 1. Remove mean.
+    mu1 = source_points.mean(axis=1, keepdims=True)
```
innerlee, line 35: double check that this is the average of three numbers `(x, y, z)` for each point

innerlee: i guess, either the ` source_points (np.ndarray([N, 3]))` is not true, or the code is wrong

innerlee: okay it transposes the data in the beginning

mmpose/core/evaluation/mesh_eval.py
```diff
@@ -0,0 +1,68 @@
...
+    Args:
+        source_points (np.ndarray([N, 3])): Source point set.
+        target_points (np.ndarray([N, 3])): Target point set.
+
+    Returns:
+        source_points_hat (np.ndarray([N, 3])): Transformed source point set.
+    """
+
+    transposed = False
+    if source_points.shape[0] != 3 and source_points.shape[0] != 2:
```
innerlee, line 28: determine whether the data is transposed or not based on this heuristic is dangerous.
it is useless when there are only two or three points.

I suggest that we remove this guess and make it clear that the data shape should be `(N, 3)`

mmpose/core/evaluation/mesh_eval.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # 2. Compute variance of X1 used for scale.
+    var1 = np.sum(X1**2)
+
+    # 3. The outer product of X1 and X2.
+    K = X1.dot(X2.T)
+
+    # 4. Solution that Maximizes trace(R'K) is R=U*V', where U, V are
+    # singular vectors of K.
+    U, s, Vh = np.linalg.svd(K)
```
innerlee, line 48: ` U, _, Vh = np.linalg.svd(K)`

innerlee:
Two minor comments and then good to go

innerlee (approved):

PR #161, opened 2020-09-28
Add build_optimizers

zengwang430521:
Add the function to bulid optimizers for sub-models in a model. This function is used in adversarial training. The function and test code are copied from mmediting.

### DIFF
--- a/mmpose/core/optimizer/builder.py
+++ b/mmpose/core/optimizer/builder.py
@@ -1,0 +1,62 @@
+# ------------------------------------------------------------------------------
+# Copied from https://github.com/open-mmlab/mmediting
+# Original licence: Copyright (c) MMEditing Authors,
+# under the Apache License 2.0
+# ------------------------------------------------------------------------------
+
+from mmcv.runner import build_optimizer
+
+
+def build_optimizers(model, cfgs):
+    """Build multiple optimizers from configs.
+
+    If `cfgs` contains several dicts for optimizers, then a dict for each
+    constructed optimizers will be returned.
+    If `cfgs` only contains one optimizer config, the constructed optimizer
+    itself will be returned.
+
+    For example,
+
+    1) Multiple optimizer configs:
+
+    .. code-block:: python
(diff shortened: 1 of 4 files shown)


tests/test_optimizer.py
```diff
@@ -0,0 +1,106 @@
...
+import torch
+import torch.nn as nn
+
+from mmpose.core import build_optimizers
+
+
+class ExampleModel(nn.Module):
+
+    def __init__(self):
+        super(ExampleModel, self).__init__()
```
innerlee, line 16: `super().__init__()`

tests/test_optimizer.py
```diff
@@ -0,0 +1,106 @@
+# ------------------------------------------------------------------------------
+# Copied from https://github.com/open-mmlab/mmediting
```
innerlee, line 2: although it is reasonable and okay, not sure if it is a good idea to start this notice within the mm-series.

innerlee: cc. @hellock , we can borrow code freely in mm-*, right?

zengwang430521: OK, I will drop this notice.

mmpose/core/optimizer/builder.py
```diff
@@ -0,0 +1,56 @@
+from mmcv.runner import build_optimizer
+
+
+def build_optimizers(model, cfgs):
+    """Build multiple optimizers from configs.
```
innerlee, line 5: is this needed because of "multiple optimizers"?

zengwang430521: Yes. The training of HMR model involves adversarial training and this needs individual optimizers for generator and discrinimator.

innerlee (approved):

tests/test_optimizer.py
hellock: Yes we can use codes from other codebases in OpenMMLab without this notice.

PR #167, opened 2020-10-01
Add mesh loss and gan loss for human mesh estimation.

zengwang430521:
Add mesh loss and gan loss for human mesh estimation.

### DIFF
--- a/mmpose/models/losses/mesh_loss.py
+++ b/mmpose/models/losses/mesh_loss.py
@@ -234,10 +234,10 @@ class MeshLoss(nn.Module):
         pred_joints_2d = self.project_points(pred_joints_3d, pred_camera)
 
         # Normalize keypoints to [-1,1]
-        # The ordinate origin of pred_joints_2d is
+        # The coordinate origin of pred_joints_2d is
         #  the center of the input image.
         pred_joints_2d = pred_joints_2d / (self.img_res / 2.)
-        # The ordinate origin of gt_joints_2d is
+        # The coordinate origin of gt_joints_2d is
         # the top left corner of the input image.
         gt_joints_2d = gt_joints_2d / (self.img_res / 2.) - 1
         loss_joints_2d = self.joints_2d_loss(pred_joints_2d, gt_joints_2d,
--- a/mmpose/models/losses/mesh_loss.py
+++ b/mmpose/models/losses/mesh_loss.py
@@ -21,7 +21,7 @@ def perspective_projection(points, rotation, translation, focal_length,
         camera_center (Tensor([B, 2])): Camera center
 
     Returns:
-        projected_points (Tensor([B, N, 3])): Projected 2D
+        projected_points (Tensor([B, N, 2])): Projected 2D
             points in image space.
     """
 
@@ -42,7 +42,6 @@ def perspective_projection(points, rotation, translation, focal_length,
     # Apply camera intrinsics
     projected_points = torch.einsum('bij,bkj->bki', K, projected_points)
     projected_points = projected_points[:, :, :-1]
-
     return projected_points
 
 
@@ -54,23 +53,23 @@ class MeshLoss(nn.Module):
     Args:
         joints_2d_loss_weight (float): Weight for loss on 2D joints.
         joints_3d_loss_weight (float): Weight for loss on 3D joints.
-        vertices_loss_weight (float): Weight for loss on 3D verteices.
+        vertex_loss_weight (float): Weight for loss on 3D verteices.
         smpl_pose_loss_weight (float): Weight for loss on SMPL
             pose parameters.
         smpl_beta_loss_weight (float): Weight for loss on SMPL
             shape parameters.
+        img_res (int): Input image resolution.
         focal_length (float): Focal length of camera model. Default=5000.
-        img_res (int): Input image resolution. Default=256.
     """
 
     def __init__(self,
                  joints_2d_loss_weight,
                  joints_3d_loss_weight,
-                 vertices_loss_weight,
+                 vertex_loss_weight,
                  smpl_pose_loss_weight,
                  smpl_beta_loss_weight,
-                 focal_length=5000,
-                 img_res=256):
+                 img_res,
+                 focal_length=5000):
 
         super().__init__()
         # Per-vertex loss on the mesh
@@ -85,7 +84,7 @@ class MeshLoss(nn.Module):
 
         self.joints_2d_loss_weight = joints_2d_loss_weight
         self.joints_3d_loss_weight = joints_3d_loss_weight
-        self.vertices_loss_weight = vertices_loss_weight
+        self.vertex_loss_weight = vertex_loss_weight
         self.smpl_pose_loss_weight = smpl_pose_loss_weight
         self.smpl_beta_loss_weight = smpl_beta_loss_weight
         self.focal_length = focal_length
@@ -120,16 +119,16 @@ class MeshLoss(nn.Module):
         else:
             return pred_joints_3d.sum() * 0
 
-    def vertices_loss(self, pred_vertices, gt_vertices, has_smpl):
+    def vertex_loss(self, pred_vertices, gt_vertices, has_smpl):
         """Compute 3D vertex loss for the examples that 3D human mesh
         annotations are available.
 
         The loss is weighted by the has_smpl.
         """
         conf = has_smpl.float()
-        loss_vertices = self.criterion_vertex(pred_vertices, gt_vertices)
-        loss_vertices = (conf[:, None, None] * loss_vertices).mean()
-        return loss_vertices
+        loss_vertex = self.criterion_vertex(pred_vertices, gt_vertices)
+        loss_vertex = (conf[:, None, None] * loss_vertex).mean()
+        return loss_vertex
 
     def smpl_losses(self, pred_rotmat, pred_betas, gt_pose, gt_betas,
                     has_smpl):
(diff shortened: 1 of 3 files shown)


mmpose/models/losses/mesh_loss.py
```diff
@@ -0,0 +1,339 @@
...
+    # Transform points
+    points = torch.einsum('bij,bkj->bki', rotation, points)
+    points = points + translation.unsqueeze(1)
+
+    # Apply perspective distortion
+    projected_points = points / points[:, :, -1].unsqueeze(-1)
+
+    # Apply camera intrinsics
+    projected_points = torch.einsum('bij,bkj->bki', K, projected_points)
+    projected_points = projected_points[:, :, :-1]
```
innerlee, line 44: what size is the last dim of projected_points

mmpose/models/losses/mesh_loss.py
```diff
@@ -0,0 +1,339 @@
...
+                                             joints_2d_visible)
+        losses['joints_2d_loss'] = loss_joints_2d * self.joints_2d_loss_weight
+
+        return losses
+
+
+@LOSSES.register_module()
+class GANLoss(nn.Module):
+    """Define GAN loss. Adapted from https://github.com/open-mmlab/mmediting
+    Original licence:  Copyright (c) MMEditing Authors, under the Apache
```
innerlee, line 251: rm notice for mm-sereis

mmpose/models/losses/mesh_loss.py
zengwang430521: It's the z coordinates. In other words, it's the depth.

mmpose/models/losses/mesh_loss.py
zengwang430521: OK, get it.

mmpose/models/losses/mesh_loss.py
innerlee: i mean, in L24, it says the returned shape is `[B, N, 3]`. but i guess the `[:, :, :-1]` might return some shape of `B, N, 2`

mmpose/models/losses/mesh_loss.py
```diff
@@ -0,0 +1,337 @@
...
+        focal_length (Tensor([B,])): Focal length
+        camera_center (Tensor([B, 2])): Camera center
+
+    Returns:
+        projected_points (Tensor([B, N, 3])): Projected 2D
+            points in image space.
+    """
+
+    batch_size = points.shape[0]
+    K = torch.zeros([batch_size, 3, 3], device=points.device)
```
innerlee, line 29: since K is the same for all samples in a batch, maybe try this

```python
K = torch.zeros([3, 3], device=points.device)
...
projected_points = torch.einsum('ij,bkj->bki', K, projected_points)
```

mmpose/models/losses/mesh_loss.py
```diff
@@ -0,0 +1,337 @@
...
+
+        # Compute 2D reprojection loss for the 2D joints
+        pred_camera = output['camera']
+        gt_joints_2d = target['joints_2d']
+        joints_2d_visible = target['joints_2d_visible']
+        pred_joints_2d = self.project_points(pred_joints_3d, pred_camera)
+
+        # Normalize keypoints to [-1,1]
+        pred_joints_2d = pred_joints_2d / (self.img_res / 2.)
+        gt_joints_2d = gt_joints_2d / (self.img_res / 2.) - 1
```
innerlee, line 240: why computation of the above two lines are different?

mmpose/models/losses/mesh_loss.py
```diff
@@ -0,0 +1,337 @@
...
+    """
+
+    def __init__(self,
+                 joints_2d_loss_weight,
+                 joints_3d_loss_weight,
+                 vertices_loss_weight,
+                 smpl_pose_loss_weight,
+                 smpl_beta_loss_weight,
+                 focal_length=5000,
+                 img_res=256):
```
innerlee, line 73: should this be required? is it possible that a user used a larger resolution and forgot to specify this?

mmpose/models/losses/mesh_loss.py
```diff
@@ -0,0 +1,337 @@
...
+        """
+        batch_size = points_3d.shape[0]
+        device = points_3d.device
+        cam_t = torch.stack([
+            camera[:, 1], camera[:, 2], 2 * self.focal_length /
+            (self.img_res * camera[:, 0] + 1e-9)
+        ],
+                            dim=-1)
+        camera_center = camera.new_zeros([batch_size, 2])
+        rot_t = torch.eye(
```
innerlee, line 174: why rot_t is hard-coded as identity? if so, then why it is needed?

mmpose/models/losses/mesh_loss.py
```diff
@@ -0,0 +1,337 @@
...
+            losses (dict): dict of losses.
+        """
+        losses = {}
+
+        # Per-vertex loss for the shape
+        pred_vertices = output['vertices']
+
+        gt_vertices = target['vertices']
+        has_smpl = target['has_smpl']
+        loss_vertices = self.vertices_loss(pred_vertices, gt_vertices,
```
innerlee, line 207: the pronunciation of `vertices_loss` is strange. how about change it to `vertex_loss`

mmpose/models/losses/mesh_loss.py
```diff
@@ -0,0 +1,337 @@
...
+            Note that loss_weight is only for generators; and it is always 1.0
+            for discriminators.
+    """
+
+    def __init__(self,
+                 gan_type,
+                 real_label_val=1.0,
+                 fake_label_val=0.0,
+                 loss_weight=1.0):
+        super(GANLoss, self).__init__()
```
innerlee, line 266: super

mmpose/models/losses/mesh_loss.py
zengwang430521: Sorry, it's a typo.  The returned shape should be [B, N, 2]. I will fix it.

mmpose/models/losses/mesh_loss.py
zengwang430521: K can be different for samples in a batch, if the focal_length and camera_center of samples are different.
So I prefer to keep K in shape [batch_size, 3, 3].

mmpose/models/losses/mesh_loss.py
zengwang430521: The ordinate origin of ` gt_joints_2d `is the top left corner of the input image. But the ordinate origin of ` pred_joints_2d` is the center of the input image. So the computation is different.

innerlee: okay, add this info as a comment line to help future readers

mmpose/models/losses/mesh_loss.py
innerlee: yeah you are right

mmpose/models/losses/mesh_loss.py
zengwang430521: OK, I wiil make the img_res required.

mmpose/models/losses/mesh_loss.py
zengwang430521: The `perspective_projection` is used to realize the projection of a perspective camera model, which includes the rotation of camera. 
However, HMR only uses a weak perspective camera model and it ignores the rotation of camera. So the `rot_t` is fixed as identity.

mmpose/models/losses/mesh_loss.py
zengwang430521: Ok , I will modify it.

mmpose/models/losses/mesh_loss.py
zengwang430521: OK , fixed

mmpose/models/losses/mesh_loss.py
innerlee: is there any model that uses "non-weak" camera model?

mmpose/models/losses/mesh_loss.py
zengwang430521: OK, I will add the comment

mmpose/models/losses/mesh_loss.py
zengwang430521: No, most models use the weak perspective projection.

innerlee: okay, let's leave it here for future reference.

mmpose/models/losses/mesh_loss.py
```diff
@@ -0,0 +1,339 @@
...
+        losses['joints_3d_loss'] = loss_joints_3d * self.joints_3d_loss_weight
+
+        # Compute 2D reprojection loss for the 2D joints
+        pred_camera = output['camera']
+        gt_joints_2d = target['joints_2d']
+        joints_2d_visible = target['joints_2d_visible']
+        pred_joints_2d = self.project_points(pred_joints_3d, pred_camera)
+
+        # Normalize keypoints to [-1,1]
+        # The ordinate origin of pred_joints_2d is
```
innerlee, line 237: coordinate origin?

mmpose/models/losses/mesh_loss.py
```diff
@@ -0,0 +1,339 @@
...
+        #  the center of the input image.
+        pred_joints_2d = pred_joints_2d / (self.img_res / 2.)
```
innerlee, line 239: just curious, 
> The ordinate origin of pred_joints_2d is the center of the input image.
if img width is 3, then what is the center of the input image? 1 or 1.5?

* * *
0 1 2

if center = 1.5, then have you checked that this convention is consistently used everywhere?
if center = 1, then its distance to the right-most pixel is 2.

zengwang430521: It should be 1.
Maybe, during the normalization, I should use `self.img_res - 1` to replace `self.img_res`.

mmpose/models/losses/mesh_loss.py
zengwang430521: I will fix it.

mmpose/models/losses/mesh_loss.py
innerlee: will you do this in a following pr, or hold this pr for a decision?

zengwang430521: I have modified the codes in this PR.

innerlee (approved):

innerlee:
Need a following pr to update the change log

PR #184, opened 2020-10-08
Add ParametricMesh detector.

zengwang430521:


requirements/runtime.txt
```diff
@@ -1,3 +1,5 @@
+chumpy
+dataclasses
```
innerlee, line 2: is this pkg used?

mmpose/models/detectors/mesh.py
```diff
@@ -0,0 +1,298 @@
...
+            real_score = self.disc(real_data)
+
+            disc_losses = {}
+            disc_losses['real_loss'] = self.loss_gan(
+                real_score, target_is_real=True, is_disc=True)
+            disc_losses['fake_loss'] = self.loss_gan(
+                fake_score, target_is_real=False, is_disc=True)
+            loss_disc, log_vars_d = self._parse_losses(disc_losses)
+
+            optimizer['disc'].zero_grad()
```
innerlee, line 137: note that generator uses the full name

requirements/runtime.txt
zengwang430521: `dataclasses` is used by smplx package, if we remove it, the test will fail.
`chumpy` is used when we use the official SMPL weight file. When using the generated test weight file, it is not used. 
If we preprocess the official SMPL weight file, we can remove the chumpy element in that file. But the preprocess function also need to use chumpy package.

mmpose/models/detectors/mesh.py
zengwang430521: OK, I will also use the full name for discriminator.

requirements/runtime.txt
innerlee: * when you install `smplx`, so it will not install `dataclass`? does this mean that it is an issue with `smplx` package?
* `chumpy is used when we use the official SMPL weight file.` do we have code that do this? i didn't find any `import chumpy`

mmpose/models/detectors/mesh.py
```diff
@@ -0,0 +1,298 @@
...
+            pose2rot=False)
+        pred_vertices = pred_out.vertices
+        pred_joints_3d = self.get_3d_joints_from_mesh(pred_vertices)
+
+        all_preds = (pred_joints_3d.detach().cpu().numpy(),
+                     (pred_pose.detach().cpu().numpy(),
+                      pred_beta.detach().cpu().numpy()),
+                     pred_camera.detach().cpu().numpy())
+
+        all_boxes = np.zeros((1, 6))
```
innerlee, line 243: dtype

requirements/runtime.txt
zengwang430521: - Because of the license of SMPL model, we can not distribute the official SMPL weight file. So in the test code, I generate a random weight file to initialize SMPL model, and the random weight file does not include chumpy elements. But in the training process, we need to use the official SMPL weight, which includes some elements need to be processed by chumpy package. Maybe I can add the dependency on chumpy in later PR.

- Under some setting, the dataclasses package is installed together with smplx,. But under some other setting, this package is not installed. I don't know the reason. Is there any possible solution?

mmpose/models/detectors/mesh.py
zengwang430521: OK, fixed.

requirements/runtime.txt
innerlee: Yes add `chumpy` when it is actually used in codes.

Sounds like an upstream issue. Here is a possible solution:
we do not deal with dataclasses directly, and redirect all related error reports to the smplx repo.
we add an faq item somewhere in the docs, to help user diagnose&solve the problem related to missing dataclasses

btw, according https://github.com/ericvsmith/dataclasses , dataclasses supports >= python3.6, and will be included in python 3.7.
On the one hand, i'm not sure whether it works on python 3.5. On the other hand, it will be included in a newer python anyway.

zengwang430521: As you see, if I remove the dataclasses in the requirements, then some checks will fail. What should I do to solve this problem?

mmpose/models/detectors/mesh.py
```diff
@@ -0,0 +1,298 @@
...
+        """Defines the computation performed at every call when testing."""
+        assert img.size(0) == 1
+        assert len(img_metas) == 1
+
+        pred_smpl = self.generator(img)
+        pred_pose, pred_beta, pred_camera = pred_smpl
+        pred_out = self.smpl(
+            betas=pred_beta,
+            body_pose=pred_pose[:, 1:],
+            global_orient=pred_pose[:, [0]],
```
innerlee, line 233: `pred_pose[:, [0]]` will make a copy, and `pred_pose[:, 0:1]` is a slice without copy.

innerlee:
looks good

mmpose/models/detectors/mesh.py
zengwang430521: OK, I will modify it to` pred_pose[:, 0].unsqueeze(1)`

innerlee: :1 is good enough

innerlee (approved):

PR #187, opened 2020-10-11
Human mesh train

zengwang430521:
Add the adversarial training code for human mesh estimation task.


mmpose/apis/train_adv.py
```diff
@@ -0,0 +1,109 @@
...
+        runner.register_hook(DistSamplerSeedHook())
+
+    # register eval hooks
+    if validate:
+        eval_cfg = cfg.get('evaluation', {})
+        val_dataset = build_dataset(cfg.data.val, dict(test_mode=True))
+        dataloader_setting = dict(
+            # samples_per_gpu=cfg.data.get('samples_per_gpu', {}),
+            samples_per_gpu=1,
+            workers_per_gpu=cfg.data.get('workers_per_gpu', {}),
```
innerlee, line 93: does workers_per_gpu accept the dict type?

mmpose/core/distributed_wrapper.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+
+    def __init__(self,
+                 module,
+                 device_ids,
+                 dim=0,
+                 broadcast_buffers=False,
+                 find_unused_parameters=False,
+                 **kwargs):
+        super(DistributedDataParallelWrapper, self).__init__()
```
innerlee, line 56: super

mmpose/datasets/datasets/mesh/mesh_h36m_dataset.py
```diff
@@ -27,7 +27,7 @@ class MeshH36MDataset(MeshBaseDataset):
             validation dataset. Default: False.
     """
 
-    def evaluate(self, outputs, res_folder, metric='joint_error'):
+    def evaluate(self, outputs, res_folder, metric='joint_error', **kwargs):
```
innerlee, line 30: where is this kwargs used

mmpose/apis/train_adv.py
zengwang430521: Actually, the type should be int. I just copy the code from mmpose/apis/train.py and do not modify this part.

mmpose/core/distributed_wrapper.py
zengwang430521: OK

mmpose/datasets/datasets/mesh/mesh_h36m_dataset.py
zengwang430521: If I remove **kwargs, it will cause an error in the eval hook. The eval hook use this function by:
`        eval_res = self.dataloader.dataset.evaluate(
            results, runner.work_dir, logger=runner.logger, **self.eval_kwargs)`
`logger` is passed into the evaluate function.

mmpose/apis/train_adv.py
innerlee: emmm, smells like a bug

zengwang430521: OK, I will change the default value to 0.

mmpose/datasets/datasets/mesh/mesh_h36m_dataset.py
innerlee: this need a better solution. pls wait

zengwang430521: OK

innerlee:
1. haven't looked into the detail, is it possible to merge the train_adv.py into train.py?
2. need a readme at ` configs/mesh/hmr/README.md`, in which the instruction to run train_adv.py should be provided.

zengwang430521:
> 1. haven't looked into the detail, is it possible to merge the train_adv.py into train.py?
> 2. need a readme at ` configs/mesh/hmr/README.md`, in which the instruction to run train_adv.py should be provided.

1. There are some little differences between train.py and train_adv.py: the adversarial dataset, multiple optimizers and no optimize hook in runner. So it's not easy to merge them.
2. I will add the readme file in further PR.
3. I have another question: I do not know how to write the test code for train_adv.py. There is not test code for train.py and I can not find the reference.

mmpose/datasets/datasets/mesh/mesh_h36m_dataset.py
innerlee: solution: put `logger=None` at the the argument list

innerlee:
as discussed with @hellock , it is better to to merge the train_adv.py into train.py. mmediting could be a reference.

mmpose/datasets/datasets/mesh/mesh_h36m_dataset.py
zengwang430521: OK

PR #188, opened 2020-10-12
return empty list, if input is empty

jin-s13:


jin-s13:
Just for safety & code readability. 
In fact, Numpy will automatically deal with np.array([]). 
Even if the input is empty, we will not get index errors.

mmpose/apis/inference.py
```diff
@@ -52,6 +52,8 @@ def _xyxy2xywh(bbox_xyxy):
         np.ndarray: Bounding boxes (with scores),
           shaped (n, 4) or (n, 5). (left, top, width, height, [score])
     """
+    if len(bbox_xyxy) == 0:
+        return []
```
innerlee, line 56: but the docstring says the return type is  np.ndarray

innerlee: this will change the returned type from np array to list, so it is not ideal.

if the purpose of the change is to speed up things, then you can skip the two lines after copy:

```
if len(bbox_xyxy) > 0:
    bbox_xywh[:, 2] = bbox_xywh[:, 2] - bbox_xywh[:, 0] + 1
    bbox_xywh[:, 3] = bbox_xywh[:, 3] - bbox_xywh[:, 1] + 1
```

PR #192, opened 2020-10-14
Add adversarial train code for human mesh estimation.

zengwang430521:


mmpose/apis/train.py
```diff
@@ -46,22 +48,39 @@ def train_model(model,
...
         model = MMDataParallel(
             model.cuda(cfg.gpu_ids[0]), device_ids=cfg.gpu_ids)
 
     # build runner
-    optimizer = build_optimizer(model, cfg.optimizer)
+    if use_adverserial_train:
+        # build multiple optimizers for generator and discriminator
+        optimizer = build_optimizers(model, cfg.optimizer)
+    else:
+        optimizer = build_optimizer(model, cfg.optimizer)
```
innerlee, line 82: can `build_optimizers` fall back on `build_optimizer`?

innerlee:
btw change these lines of code https://github.com/open-mmlab/mmpose/blob/master/mmpose/core/optimizer/builder.py#L45-L49 to 
```python
if all(isinstance(v, dict) for v in cfgs.values()):
```

mmpose/apis/train.py
zengwang430521: In adversarial training, we need to build multiple optimizers, so I use the function `build_optimizers` in mmediting.  Maybe we can use `build_optimizers` to replace `build_optimizer`.

innerlee (approved):

PR #196, opened 2020-10-17
detector return heatmaps

jin-s13:


mmpose/datasets/datasets/bottom_up/bottom_up_coco.py
```diff
@@ -204,7 +204,7 @@ def evaluate(self, outputs, res_folder, metric='mAP', **kwargs):
             num_keypoints: K
 
         Args:
-            outputs (list(preds, scores, image_path)):Output results.
+            outputs (list(preds, scores, image_path, output_cpu)):
```
innerlee, line 207: could `output_cpu` be more specific/concrete? say, things like `heatmap` etc.

mmpose/models/detectors/bottom_up.py
```diff
@@ -243,7 +243,9 @@ def forward_test(self, img, img_metas, **kwargs):
         image_path = []
         image_path.extend(img_metas['image_file'])
 
-        return results, scores, image_path
+        output_cpu = aggregated_heatmaps.clone().cpu().numpy()
```
innerlee, line 246: should clone -> detach?

innerlee:
the file `tests/data/smpl/test_joint_regressor.npy` is too large. try to use a smaller test data size

jin-s13:
> the file `tests/data/smpl/test_joint_regressor.npy` is too large. try to use a smaller test data size

maybe not related to this PR

mmpose/datasets/datasets/bottom_up/bottom_up_coco.py
jin-s13: changed to `output_heatmap`

mmpose/models/detectors/bottom_up.py
jin-s13: Yes. should be `detach()`

jin-s13:
> the file `tests/data/smpl/test_joint_regressor.npy` is too large. try to use a smaller test data size

Yes. This file should not be uploaded. I will delete it.

innerlee (approved):

PR #200, opened 2020-10-18
Use fp16 utils from mmcv

### LINKED ISSUE #153: Use fp16 utils from mmcv
ref

https://github.com/open-mmlab/mmaction2/pull/25/files

https://github.com/open-mmlab/mmaction2/pull/204/files

jin-s13:
Fix #153


innerlee:
I haven't got the chance to try fp16, so @hellock may review this

hellock:
How about the benchmark?

jin-s13:
Still have some problems to solve:
1. The training is not stable...  The accuracy is normal at first, however, it will suddenly decline to 0 at some point (about 140 epochs).
2. The results may contain NAN values. 
3. Currently, the loss calculation & post-processing use fp32.

PR #203, opened 2020-10-20
return heatmaps only if needed

liuxin9608:
to save memory, do not return heatmaps by default.

### DIFF
--- a/mmpose/models/detectors/bottom_up.py
+++ b/mmpose/models/detectors/bottom_up.py
@@ -245,7 +245,11 @@ class BottomUp(BasePose):
 
         output_heatmap = aggregated_heatmaps.detach().cpu().numpy()
 
-        return results, scores, image_path, output_heatmap
+        if 'return_heatmap' in self.test_cfg and self.test_cfg[
+                'return_heatmap']:
+            return results, scores, image_path, output_heatmap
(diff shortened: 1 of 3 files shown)


mmpose/models/detectors/bottom_up.py
```diff
@@ -245,7 +245,11 @@ def forward_test(self, img, img_metas, **kwargs):
 
         output_heatmap = aggregated_heatmaps.detach().cpu().numpy()
```
innerlee, line 246: why compute it when they are not needed?

mmpose/models/detectors/top_down.py
```diff
@@ -207,7 +207,11 @@ def forward_test(self, img, img_metas, **kwargs):
         all_boxes[0, 5] = score
         image_path.extend(img_metas['image_file'])
 
-        return all_preds, all_boxes, image_path, output_heatmap
+        if 'return_heatmap' in self.test_cfg and self.test_cfg[
```
innerlee, line 210: How about put `return_heatmap=False` in the signature of the `forward_test()`?

innerlee: and remember to change the demo code accordingly

Issue #186, opened 2020-10-11
Support different result formats of mmdet in demo


RSKothari:
```
bbox_xywh[:, 2] = bbox_xywh[:, 2] - bbox_xywh[:, 0] + 1
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```

The following script is used to call mmpose on a video
`
python $PATH_MMPOSE/demo/top_down_video_demo_with_mmdet.py $PATH_MMDET/configs/cascade_rcnn/cascade_mask_rcnn_x101_32x4d_fpn_20e_coco.py $PATH_WEIGHTS/MMDet/cascade_mask_rcnn_x101_32x4d_fpn_20e_coco_20200528_083917-ed1f4751.pth $PATH_MMPOSE/configs/top_down/darkpose/coco-wholebody/hrnet_w48_coco_wholebody_384x288_dark_plus.py $PATH_WEIGHTS/MMPose/hrnet_w48_coco_wholebody_384x288_dark-f5726563_20200918.pth --video-path test_video.mp4 --out-video-root ./results --show`

RSKothari:
A print of the variable revealed that the bounding boxes were almost completely empty


```
[array([[221.10242   , 285.07632   , 462.33337   , 713.4651    ,
          0.9974692 ],
       [559.7312    , 416.71567   , 728.47736   , 714.90625   ,
          0.99740857]], dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
 array([], shape=(0, 5), dtype=float32)
[...]

PR #217, opened 2020-10-26
support different return formats of mmdet

jin-s13:
#186


innerlee (approved):

innerlee:
Need to write in the style of `Fix #186`, so that github will create auto link. details see https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#about-issue-references

Issue #201, opened 2020-10-20
Extra memory consumption in training process


HuangJunJie2017:
![image](https://user-images.githubusercontent.com/31001081/96533681-e16d7580-12c0-11eb-9af5-6f7a39d59b0f.png)
In training, there are 8 extra processes who occupy memory of the same one gpu. This limits the batch size of the training process.

jin-s13:
It seems that you are using `dist_train.sh` to train the models. These processes are the DataLoader workers.

It is highly recommended to use `slurm_train.sh` instead of `dist_train.sh`, even in single-machine training settings. `slurm_train` use `DistributedDataParallel` which is much more efficient than `DataParallel`.

HuangJunJie2017:
@jin-s13 ok

HuangJunJie2017:
https://github.com/open-mmlab/mmpose/blob/master/mmpose/models/backbones/utils/utils.py 

map_location= 'cpu' can not help avoiding the extra gpu memory consumption when using 

>pretrained='https://download.openmmlab.com/mmpose/pretrain_models/hrnet_w32-36af842e.pth',

it works only when I downloaded the Pretrained Model and used a local address:

>pretrained="/mnt/cephfs/algorithm/junjie.huang/models/mmpose/hrnet_w32-36af842e.pth",

emmm, amazing

innerlee:
Looks like an mmcv bug.

PR #227, opened 2020-11-03
wipe out the extra gpu memony cost in dist_train

HuangJunJie2017:
wipe out the extra gpu memony cost in dist_train by set the default map_location as 'cpu' , issue#201 https://github.com/open-mmlab/mmpose/issues/201


jin-s13:
Thanks! 
Please sign CLA, so that we can merge this pull request.

innerlee (approved):

PR #288, opened 2020-11-20
add new post-process

wusize:
Migration guide:

From 
```yaml
        post_process=True,
        unbiased_decoding=False,
```
to
```yaml
        post_process='default',
```

---

, and from
```yaml
        post_process=True,
        unbiased_decoding=True,
```
to
```yaml
        post_process='unbiased',
```

---

, and from
```yaml
        post_process=False,
```
to
```yaml
        post_process=None,
```

### DIFF
--- a/configs/top_down/rsn/coco/single_ctf_rsn18_coco_256x192.py
+++ b/configs/top_down/rsn/coco/single_ctf_rsn18_coco_256x192.py
@@ -61,7 +61,6 @@ model = dict(
         post_process=True,
         shift_heatmap=False,
         unbiased_decoding=False,
-        megvii_decoding=True,
         modulate_kernel=5),
     loss_pose=[dict(type='JointsMSELoss', use_target_weight=True)] * 3 +
     [dict(type='JointsOHKMMSELoss', use_target_weight=True)])
--- a/configs/top_down/rsn/coco/single_ctf_rsn18_coco_256x192.py
+++ b/configs/top_down/rsn/coco/single_ctf_rsn18_coco_256x192.py
@@ -58,9 +58,8 @@ model = dict(
     train_cfg=dict(num_units=4, loss_weights=[0.25] * 3 + [1]),
     test_cfg=dict(
         flip_test=True,
-        post_process=True,
+        post_process='default',
         shift_heatmap=False,
-        unbiased_decoding=False,
         modulate_kernel=5),
     loss_pose=[dict(type='JointsMSELoss', use_target_weight=True)] * 3 +
     [dict(type='JointsOHKMMSELoss', use_target_weight=True)])
--- a/configs/top_down/rsn/coco/single_ctf_rsn18_coco_256x192.py
+++ b/configs/top_down/rsn/coco/single_ctf_rsn18_coco_256x192.py
@@ -59,9 +59,10 @@ model = dict(
     test_cfg=dict(
         flip_test=True,
         post_process=True,
-        shift_heatmap=True,
+        shift_heatmap=False,
         unbiased_decoding=False,
-        modulate_kernel=11),
+        megvii_decoding=True,
+        modulate_kernel=5),
     loss_pose=[dict(type='JointsMSELoss', use_target_weight=True)] * 3 +
     [dict(type='JointsOHKMMSELoss', use_target_weight=True)])
 
--- a/mmpose/core/evaluation/top_down_eval.py
+++ b/mmpose/core/evaluation/top_down_eval.py
@@ -332,7 +332,7 @@ def keypoints_from_heatmaps(heatmaps,
         - maxvals (np.ndarray[N, K, 1]): Scores (confidence) of the keypoints.
     """
 
-    assert not megvii and unbiased
+    assert not (megvii and unbiased)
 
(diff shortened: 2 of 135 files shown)


wusize:
Add post-process for mspn and rsn.

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -313,9 +314,13 @@ def keypoints_from_heatmaps(heatmaps,
         scale (np.ndarray[N, 2]): Scale of the bounding box
             wrt height/width.
         post_process (bool): Option to use post processing or not.
-        unbiased (bool): Option to use unbiased decoding.
+        unbiased (bool): Option to use unbiased decoding. Mutually
+            exclusive with megvii.
             Paper ref: Zhang et al. Distribution-Aware Coordinate
             Representation for Human Pose Estimation (CVPR 2020).
+        megvii (bool):  Option to use megvii's way of post processing.
```
innerlee, line 321: for mutually exclusive options, could they be merged into a selective option, something like `post_process=None|unbiased|megvii`?

wusize: Yes, but if we do it this way, we need to change all the config files.

innerlee: We can change them in a backward-compatible way. We add the new way of `post_process` option while keep the old way functional, and emit an deprecation warning for the old ways.

innerlee: With detailed and friendly migration message, I believe it will be a smooth transition

wusize:
All checks have passed.

configs/top_down/rsn/coco/single_ctf_rsn18_coco_256x192.py
```diff
@@ -59,9 +59,9 @@
     test_cfg=dict(
         flip_test=True,
         post_process=True,
-        shift_heatmap=True,
+        shift_heatmap=False,
```
innerlee, line 62: Will this change affect accuracies (if any) reported in model zoo?

jin-s13: This is the config for RSN. And other models are not changed. So it will not affect the model zoo.

mmpose/models/detectors/top_down.py
```diff
@@ -225,13 +225,21 @@ def process_head(self, output, img, img_metas, return_heatmap=False):
...
+                s,
+                post_process=self.test_cfg['post_process'],
+                unbiased=self.test_cfg['unbiased_decoding'],
+                kernel=self.test_cfg['modulate_kernel'])
+        else:
+            preds, maxvals = keypoints_from_heatmaps(
+                output_heatmap,
+                c,
+                s,
+                unbiased=self.test_cfg['unbiased_decoding'],
```
jin-s13, line 241: If 'unbiased_decoding' not in self.test_cfg, do we still need this input?

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -353,9 +392,17 @@ def keypoints_from_heatmaps(heatmaps,
...
+
+    else:
+        preds, maxvals = _get_max_preds(heatmaps)
 
     # Transform back to the image
     for i in range(N):
         preds[i] = transform_preds(preds[i], center[i], scale[i], [W, H])
 
+    if post_process == 'megvii':
+        maxvals = maxvals / 255 + 0.5
```
jin-s13, line 406: `255` to `255.0`

jin-s13:
Some minor comments.

mmpose/models/detectors/top_down.py
innerlee: use defaults in this case

wusize: Sorry, this is a mistake. I erroneously removed **post_process=self.test_cfg['post_process']** while I should remove **unbiased**

mmpose/core/evaluation/top_down_eval.py
wusize: Ok

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -312,8 +314,13 @@ def keypoints_from_heatmaps(heatmaps,
         center (np.ndarray[N, 2]): Center of the bounding box (x, y).
         scale (np.ndarray[N, 2]): Scale of the bounding box
             wrt height/width.
-        post_process (bool): Option to use post processing or not.
-        unbiased (bool): Option to use unbiased decoding.
+        post_process (str/None): Choice of methods to post-process
+            heatmaps. Currently supported: None, 'default', 'unbiased',
+            'megvii'.
+        unbiased (bool): Option to use unbiased decoding. Mutually
```
innerlee, line 320: wait, so `unbiased` is new in this pr?
If so, we do not need to deprecate it.

innerlee:
The name `default` does not feels perfect. Any suggestions? @jin-s13 @wusize @Canwang-sjtu

innerlee (approved):

PR #291, opened 2020-11-22
Udp for top-down and bottom up

HuangJunJie2017:


jin-s13:
@HuangJunJie2017  Please sign CLA :)

HuangJunJie2017:
@jin-s13 i have sign the CLA,
image

configs_udp/bottom_up/higherhrnet/coco/higher_hrnet32_coco_512x512.py
```diff
@@ -0,0 +1,205 @@
+log_level = 'INFO'
```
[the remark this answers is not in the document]
jin-s13: That would be better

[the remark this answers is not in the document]
HuangJunJie2017: emmm, considering that i will contribute another performance improvement method named AID. 
The same as UDP, AID is also an model-agnostic/paradigm-agnostic approach.  So, configs/bottom_up/udp_pose/higherhrnet/coco/higher_hrnet32_coco_512x512.py for udp and configs/bottom_up/aid_pose/higherhrnet/coco/higher_hrnet32_coco_512x512.py for aid?
AID: https://arxiv.org/abs/2008.07139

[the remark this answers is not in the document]
HuangJunJie2017: How about the following arrangement?
original: configs/bottom_up/higherhrnet/coco/higher_hrnet32_coco_512x512.py
udp: configs/bottom_up/higherhrnet/coco/higher_hrnet32_coco_512x512_udp.py
aid: configs/bottom_up/higherhrnet/coco/higher_hrnet32_coco_512x512_udp_aid.py

HuangJunJie2017:
It is very strange that when i sign the CLA, the status here is still unsigned but i am noticed that i have signed here:
image
After few hours later(may be >12 hours), i can sign this again, but it still doesn't work~

configs_udp/bottom_up/higherhrnet/coco/higher_hrnet32_coco_512x512.py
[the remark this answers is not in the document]
jin-s13:
```
udp: configs/bottom_up/higherhrnet/coco/higher_hrnet32_coco_512x512_udp.py
aid: configs/bottom_up/higherhrnet/coco/higher_hrnet32_coco_512x512_udp_aid.py
```
That's better.

jin-s13:
image
Please check again if these two users are the same... `junjie.huang` and `HuangJunJie2017`.
Do not worry. We can still proceed and regard the snapshot as an agreement : )

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -327,35 +422,68 @@ def keypoints_from_heatmaps(heatmaps,
...
-            heatmaps = np.log(heatmaps)
-            for n in range(N):
-                for k in range(K):
-                    preds[n][k] = _taylor(heatmaps[n][k], preds[n][k])
+    if use_udp:
+        if target_type == 'GaussianHeatMap':
+            coords, maxvals = _get_max_preds(heatmaps)
+            if post_process:
+                coords = post_dark(coords, heatmaps)
+        elif target_type == 'CombinedTarget':
```
jin-s13, line 431: Please also add a comment to explain briefly, what is "CombinedTarget"  : )

HuangJunJie2017:
> image
> Please check again if these two users are the same... `junjie.huang` and `HuangJunJie2017`.
> Do not worry. We can still proceed and regard the snapshot as an agreement : )

Nice, that is the point, I committed with another email. I add this email to this account and sign again. It works~

mmpose/core/evaluation/bottom_up_eval.py
```diff
@@ -181,13 +197,26 @@ def get_group_preds(grouped_joints, center, scale, heatmap_size):
...
+        use_udp (bool): Unbiased data processing.
 
     Returns:
         list: List of the pose result for each person.
     """
-    results = []
-    for person in grouped_joints[0]:
-        joints = transform_preds(person, center, scale, heatmap_size)
-        results.append(joints)
+    if use_udp:
```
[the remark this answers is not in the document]
HuangJunJie2017: they are not equivalent~

configs_udp/bottom_up/higherhrnet/coco/higher_hrnet32_coco_512x512.py
```diff
@@ -0,0 +1,205 @@
...
+load_from = None
+resume_from = None
+dist_params = dict(backend='nccl')
+workflow = [('train', 1)]
+checkpoint_config = dict(interval=5)
+evaluation = dict(interval=300, metric='mAP')
```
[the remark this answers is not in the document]
HuangJunJie2017: done

configs_udp/top_down/hrnet/coco/hrnet_w32_coco_256x192.py
```diff
@@ -0,0 +1,179 @@
...
+    num_output_channels=channel_cfg['num_output_channels'],
+    num_joints=channel_cfg['dataset_joints'],
+    dataset_channel=channel_cfg['dataset_channel'],
+    inference_channel=channel_cfg['inference_channel'],
+    soft_nms=False,
+    nms_thr=1.0,
+    oks_thr=0.9,
+    vis_thr=0.2,
+    bbox_thr=1.0,
+    use_gt_bbox=True,
```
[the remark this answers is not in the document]
HuangJunJie2017: done

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,95 @@ def _taylor(heatmap, coord):
...
+    batch_heatmaps_pad[:, :, 0, 1:-1] = batch_heatmaps[:, :, 0, :]
+    batch_heatmaps_pad[:, :, -1, -1] = batch_heatmaps[:, :, -1, -1]
+    batch_heatmaps_pad[:, :, 0, 0] = batch_heatmaps[:, :, 0, 0]
+    batch_heatmaps_pad[:, :, 0, -1] = batch_heatmaps[:, :, 0, -1]
+    batch_heatmaps_pad[:, :, -1, 0] = batch_heatmaps[:, :, -1, 0]
+    i_ = np.zeros((coord_shape[0], coord_shape[1]))
+    ix1 = np.zeros((coord_shape[0], coord_shape[1]))
+    iy1 = np.zeros((coord_shape[0], coord_shape[1]))
+    ix1y1 = np.zeros((coord_shape[0], coord_shape[1]))
+    ix1_y1_ = np.zeros((coord_shape[0], coord_shape[1]))
```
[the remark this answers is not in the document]
HuangJunJie2017: different position in data matrix:
ix1_y1_ & iy1_ & *       \\
ix1_      & i       & ix1    \\
*           & iy1   & ix1y1\\

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,95 @@ def _taylor(heatmap, coord):
...
+                                           coords[i, j, 0] + 1]
+            iy1_[i, j] = batch_heatmaps_pad[0, j, coords[i, j, 1],
+                                            coords[i, j, 0] + 1]
+            ix1y1[i, j] = batch_heatmaps_pad[0, j, coords[i, j, 1] + 2,
+                                             coords[i, j, 0] + 2]
+            ix1_y1_[i, j] = batch_heatmaps_pad[0, j, coords[i, j, 1],
+                                               coords[i, j, 0]]
+    dx = 0.5 * (ix1 - ix1_)
+    dy = 0.5 * (iy1 - iy1_)
+    D = np.zeros((coord_shape[0], coord_shape[1], 2))
```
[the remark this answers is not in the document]
HuangJunJie2017: use 'first_derivative' instead?

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,95 @@ def _taylor(heatmap, coord):
...
+    for i in range(coord_shape[0]):
+        for j in range(coord_shape[1]):
+            hessian_tmp = hessian[i, j, :, :] + 1e-8 * np.eye(2)
+            inv_hessian[i, j, :, :] = np.linalg.inv(hessian_tmp)
+    res = np.zeros(coords.shape)
+    coords = coords.astype(np.float)
+    for i in range(coord_shape[0]):
+        for j in range(coord_shape[1]):
+            D_tmp = D[i, j, :]
+            D_tmp = D_tmp[:, np.newaxis]
```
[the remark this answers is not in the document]
HuangJunJie2017: done

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,95 @@ def _taylor(heatmap, coord):
...
+    dxy = 0.5 * (ix1y1 - ix1 - iy1 + i_ + i_ - ix1_ - iy1_ + ix1_y1_)
+    hessian = np.zeros((coord_shape[0], coord_shape[1], 2, 2))
+    hessian[:, :, 0, 0] = dxx
+    hessian[:, :, 1, 0] = dxy
+    hessian[:, :, 0, 1] = dxy
+    hessian[:, :, 1, 1] = dyy
+    inv_hessian = np.zeros(hessian.shape)
+    for i in range(coord_shape[0]):
+        for j in range(coord_shape[1]):
+            hessian_tmp = hessian[i, j, :, :] + 1e-8 * np.eye(2)
```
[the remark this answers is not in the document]
HuangJunJie2017: how? is there an example?

mmpose/core/evaluation/bottom_up_eval.py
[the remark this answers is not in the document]
Canwang-sjtu: why

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: i will add the following comment in the args of any functions that involve 'CombinedTarget' ~
        target_type (str): supported targets: 'GaussianHeatMap', 
            'CombinedTarget'. Default:'GaussianHeatMap'
            CombinedTarget: The combination of classification target
            (response map) and regression target (offset map).
            Paper ref: Huang et al. The Devil is in the Details: Delving into 
            Unbiased Data Processing for Human Pose Estimation (CVPR 2020).

mmpose/core/evaluation/bottom_up_eval.py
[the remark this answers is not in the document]
HuangJunJie2017: the return of affine_joints(grouped_joints[0][:, :, :2], trans) is all person keypoints position. 
in your method, we will get:
if use_udp:
     result = [ndarray[N,C,2],]
else:
     result = [ndarray[C,3],ndarray[C,3],ndarray[C,3],....]

mmpose/core/evaluation/top_down_eval.py
jin-s13: Nice.

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,95 @@ def _taylor(heatmap, coord):
...
+    """DARK post-pocessing.
+
+    Args:
+        coords (np.ndarray): Coords in shape of batchsize*num_kps*2.
+        batch_heatmaps (np.ndarray): batch_heatmaps in shape of
+                batchsize*num_kps*high*width.
+
+    Returns:
+        res (np.ndarray): Refined coords in shape of batchsize*num_kps*2.
+    """
```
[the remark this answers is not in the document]
HuangJunJie2017: done

mmpose/core/evaluation/bottom_up_eval.py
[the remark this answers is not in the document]
Canwang-sjtu: thanks for your reminder :)

mmpose/core/evaluation/top_down_eval.py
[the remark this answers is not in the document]
Canwang-sjtu: its ok

mmpose/core/evaluation/top_down_eval.py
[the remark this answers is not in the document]
Canwang-sjtu: 1e-8 is relatively arbitrary, We don’t know if it’s useful in practice. 
or you can add a simple comment?

mmpose/models/detectors/top_down.py
```diff
@@ -153,25 +154,28 @@ def forward_train(self, img, target, target_weight, img_metas, **kwargs):
...
                     target.detach().cpu().numpy(),
                     target_weight.detach().cpu().numpy().squeeze(-1) > 0)
         else:
-            _, avg_acc, _ = pose_pck_accuracy(
-                output.detach().cpu().numpy(),
-                target.detach().cpu().numpy(),
-                target_weight.detach().cpu().numpy().squeeze(-1) > 0)
-
+            avg_acc = 0.0
         losses['acc_pose'] = float(avg_acc)
```
Canwang-sjtu, line 180: if self.target_type == 'GaussianHeatMap':
            if isinstance(output, list):
                if target.dim() == 5 and target_weight.dim() == 4:
                    _, avg_acc, _ = pose_pck_accuracy(
                        output[-1].detach().cpu().numpy(),
                        target[:, -1, ...].detach().cpu().numpy(),
                        target_weight[:, -1,
                                      ...].detach().cpu().numpy().squeeze(-1) >
                        0)
                    # Only use the last output for prediction
                else:
                    _, avg_acc, _ = pose_pck_accuracy(
                        output[-1].detach().cpu().numpy(),
                        target.detach().cpu().numpy(),
                        target_weight.detach().cpu().numpy().squeeze(-1) > 0)
            else:
                _, avg_acc, _ = pose_pck_accuracy(
                    output.detach().cpu().numpy(),
                    target.detach().cpu().numpy(),
                    target_weight.detach().cpu().numpy().squeeze(-1) > 0)
            losses['acc_pose'] = float(avg_acc)

If target_type is equal to'CombinedTarget',  we do not need to output acc_pose

mmpose/core/evaluation/top_down_eval.py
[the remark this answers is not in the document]
HuangJunJie2017: It is necessary for ensuring the full rank of the matrix, and with repeated experiments it's effect is small enough.
to use 'np.finfo(np.float32).eps' instead?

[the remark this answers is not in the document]
Canwang-sjtu: np.finfo(np.float32).eps is better ~

mmpose/models/detectors/top_down.py
HuangJunJie2017: done

mmpose/datasets/pipelines/top_down_transform.py
```diff
@@ -356,12 +382,113 @@ def _megvii_generate_target(self, cfg, joints_3d, joints_3d_visible,
...
+                dtype=np.float32)
+            feat_width = heatmap_size[0]
+            feat_height = heatmap_size[1]
+            feat_x_int = np.arange(0, feat_width)
+            feat_y_int = np.arange(0, feat_height)
+            feat_x_int, feat_y_int = np.meshgrid(feat_x_int, feat_y_int)
+            feat_x_int = feat_x_int.reshape((-1, ))
+            feat_y_int = feat_y_int.reshape((-1, ))
+            kps_pos_distance_x = factor / 64.0 * heatmap_size[1]
+            kps_pos_distance_y = factor / 64.0 * heatmap_size[1]
```
Canwang-sjtu, line 462: what's the meaning of 64?

mmpose/datasets/pipelines/top_down_transform.py
```diff
@@ -356,12 +382,113 @@ def _megvii_generate_target(self, cfg, joints_3d, joints_3d_visible,
...
+        """
+        num_joints = cfg['num_joints']
+        image_size = cfg['image_size']
+        heatmap_size = cfg['heatmap_size']
+        joint_weights = cfg['joint_weights']
+        use_different_joint_weights = cfg['use_different_joint_weights']
+
+        target_weight = np.ones((num_joints, 1), dtype=np.float32)
+        target_weight[:, 0] = joints_3d_visible[:, 0]
+
```
Canwang-sjtu, line 408: delete line 478 and  479

        else:
            assert False

mmpose/datasets/pipelines/top_down_transform.py
HuangJunJie2017: when the input size is 256x192, the height of the hrnet heatmap is 64
use '0.0546875'=3.5/64.0 instead of '3.5'? in this way, we can remove this constant '64.0'

mmpose/datasets/pipelines/top_down_transform.py
```diff
@@ -356,12 +382,113 @@ def _megvii_generate_target(self, cfg, joints_3d, joints_3d_visible,
...
+            kps_pos_distance_x = factor / 64.0 * heatmap_size[1]
+            kps_pos_distance_y = factor / 64.0 * heatmap_size[1]
+            feat_stride = (image_size - 1.0) / (heatmap_size - 1.0)
+            for joint_id in range(num_joints):
+                mu_x = joints_3d[joint_id][0] / feat_stride[0]
+                mu_y = joints_3d[joint_id][1] / feat_stride[1]
+                x_offset = (mu_x - feat_x_int) / kps_pos_distance_x
+                y_offset = (mu_y - feat_y_int) / kps_pos_distance_y
+                dis = x_offset**2 + y_offset**2
+                keep_pos = np.where((dis <= 1) & (dis >= 0))[0]
```
Canwang-sjtu, line 470: dis is always >= 0

mmpose/datasets/pipelines/top_down_transform.py
```diff
@@ -403,6 +530,35 @@ def __call__(self, results):
                 target, target_weight = self._megvii_generate_target(
                     results['ann_info'], joints_3d, joints_3d_visible,
                     self.kernel)
+        elif self.encoding == 'UDP':
+            if self.target_type == 'CombinedTarget':
```
Canwang-sjtu, line 534: `assert target_type in ['GaussianHeatMap', 'CombinedTarget']`

delete line 540 and 541

mmpose/datasets/pipelines/top_down_transform.py
jin-s13: If we use different input_size (e.g. 384x288), shall we change this?

HuangJunJie2017: No~ just let it in direct proportion to the heatmap_size( or input_size)
I have searched for this hyper-parameter in both 256x192 and 384x288 input_size. Based on the experimental result, the optimal distance is in direct proportion to the heatmap_height with a scale factor of '0.0546875'.

HuangJunJie2017: however, for other tasks( e.g. face landmark, hand pose), this setting may not be the optimal one~

mmpose/models/losses/mse_loss.py
```diff
@@ -42,6 +42,48 @@ def forward(self, output, target, target_weight):
...
+    """
+
+    def __init__(self, use_target_weight):
+        super(JointsMSELoss_Combined, self).__init__()
+        self.criterion = nn.MSELoss(reduction='mean')
+        self.use_target_weight = use_target_weight
+
+    def forward(self, output, target, target_weight):
+        batch_size = output.size(0)
+        num_joints = output.size(1)
```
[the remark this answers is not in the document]
HuangJunJie2017: emm， to avoid the misunderstanding, i modify this function into this
>@LOSSES.register_module()
>class` JointsMSELoss_Combined(nn.Module):
    """MSE loss for combined target.
            CombinedTarget: The combination of classification target
            (response map) and regression target (offset map).
            Paper ref: Huang et al. The Devil is in the Details: Delving into
            Unbiased Data Processing for Human Pose Estimation (CVPR 2020).

    Args:
        use_target_weight (bool): Option to use weighted MSE loss.
            Different joint types may have different target weights.
    """

    def __init__(self, use_target_weight):
        super().__init__()
        self.criterion = nn.MSELoss(reduction='mean')
        self.use_target_weight = use_target_weight

    def forward(self, output, target, target_weight):
        batch_size, num_channels, _, _ = output.size
        heatmaps_pred = output.reshape(
            (batch_size, num_channels, -1)).split(1, 1)
        heatmaps_gt = target.reshape(
            (batch_size, num_channels, -1)).split(1, 1)
        loss = 0.
        num_joints = num_channels // 3
        for idx in range(num_joints):
            heatmap_pred = heatmaps_pred[idx * 3].squeeze()
            heatmap_gt = heatmaps_gt[idx * 3].squeeze()
            offset_x_pred = heatmaps_pred[idx * 3 + 1].squeeze()
            offset_x_gt = heatmaps_gt[idx * 3 + 1].squeeze()
            offset_y_pred = heatmaps_pred[idx * 3 + 2].squeeze()
            offset_y_gt = heatmaps_gt[idx * 3 + 2].squeeze()
            if self.use_target_weight:
                heatmap_pred = heatmap_pred.mul(target_weight[:, idx])
                heatmap_gt = heatmap_gt.mul(target_weight[:, idx])
            # classification loss
            loss += 0.5 * self.criterion(heatmap_pred, heatmap_gt)
            # regression loss
            loss += 0.5 * self.criterion(heatmap_gt * offset_x_pred,
                                         heatmap_gt * offset_x_gt)
            loss += 0.5 * self.criterion(heatmap_gt * offset_y_pred,
                                         heatmap_gt * offset_y_gt)
        return loss / num_joints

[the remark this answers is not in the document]
Canwang-sjtu: nice

innerlee:
I will take a pass on code styles this week

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+        share_heatmap (bool) : True for bottom_up paradigm where all persons
+                    share the same heatmap while False for top_down paradigm
+                    where each person has its own heatmaps.
+
+    Returns:
+        res (np.ndarray): Refined coords in shape of batchsize*num_kps*2.
+    """
+    if not isinstance(batch_heatmaps, np.ndarray):
+        batch_heatmaps = batch_heatmaps.cpu().numpy()
+    batch, num_kps, h, w = batch_heatmaps.shape
```
innerlee, line 272: `N, K, H, W = batch_heatmaps.shape`

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+def post_dark(coords, batch_heatmaps, share_heatmap=True):
+    """DARK post-pocessing.
+
+    Note:
+        batch_size: N
+        num_keypoints: K
+
+    Args:
+        coords (np.ndarray[N, K, 2]): Coords in shape of batchsize*num_kps*2.
+        batch_heatmaps (np.ndarray[N, K, H, W]): batch_heatmaps
```
innerlee, line 262: add a sentence about the "if share_heatmap then batch = 1"

innerlee: add this to the `share_heatmap (bool) : ` part is also okay

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+        batch_heatmaps = batch_heatmaps.cpu().numpy()
+    batch, num_kps, h, w = batch_heatmaps.shape
+    persons = coords.shape[0]
+    if share_heatmap:
+        assert batch == 1
+    else:
+        assert batch == persons
+    for i in range(batch):
+        for j in range(num_kps):
+            mapij = batch_heatmaps[0, j, :, :]
```
innerlee, line 280: double check you want `batch_heatmaps[0, j, :, :]` rather than `batch_heatmaps[i, j, :, :]`

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+                    share the same heatmap while False for top_down paradigm
+                    where each person has its own heatmaps.
+
+    Returns:
+        res (np.ndarray): Refined coords in shape of batchsize*num_kps*2.
+    """
+    if not isinstance(batch_heatmaps, np.ndarray):
+        batch_heatmaps = batch_heatmaps.cpu().numpy()
+    batch, num_kps, h, w = batch_heatmaps.shape
+    persons = coords.shape[0]
```
innerlee, line 273: use `B = coords.shape[0]`

Both B and N can be used to represent batch size.

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+    if share_heatmap:
+        assert batch == 1
+    else:
+        assert batch == persons
+    for i in range(batch):
+        for j in range(num_kps):
+            mapij = batch_heatmaps[0, j, :, :]
+            mapij = cv2.GaussianBlur(mapij, (3, 3), 0)
+            batch_heatmaps[0, j, :, :] = mapij
```
innerlee, line 282: cv2 generally supports inplace operations

Try
```python
    for i in range(N):
        for j in range(K):
            mapij = batch_heatmaps[i, j]
            cv2.GaussianBlur(mapij , (3, 3), 0, mapij)
```

innerlee: Or
```python
    for heatmaps in batch_heatmaps:
        for heatmap in heatmaps :
            cv2.GaussianBlur(heatmap, (3, 3), 0, heatmap)
```

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+    batch_heatmaps = np.clip(batch_heatmaps, 0.001, 50)
```
innerlee, line 283: np also support inplace operations, try
```python
np.clip(batch_heatmaps, 0.001, 50, batch_heatmaps)
```

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+    batch_heatmaps = np.log(batch_heatmaps)
+    batch_heatmaps = np.transpose(batch_heatmaps, (2, 3, 0, 1)).reshape(
+        (h, w, batch * num_kps))
```
innerlee, line 286: `.reshape(H, W, -1)`

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+    batch_heatmaps_pad = cv2.copyMakeBorder(
+        batch_heatmaps, 1, 1, 1, 1, borderType=cv2.BORDER_REFLECT)
+    batch_heatmaps_pad = np.transpose(
+        batch_heatmaps_pad.reshape((h + 2, w + 2, batch, num_kps)),
+        (2, 3, 0, 1))
+    index = coords[:, :, 0] + 1 + (coords[:, :, 1] + 1) * (w + 2)
```
innerlee, line 292: `[:, :, 0]` -> `[..., 0]`

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+
+    index = index.reshape((persons, -1))
```
innerlee, line 294: needless brackets

innerlee: btw why do you need this reshape?

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+    if share_heatmap:
+        offset = np.arange(0, num_kps) * (w + 2) * (h + 2)
+        index = index + offset[None, :]
```
innerlee, line 297: I guess the `[None, :]` could be omitted

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+        (h, w, batch * num_kps))
+    batch_heatmaps_pad = cv2.copyMakeBorder(
+        batch_heatmaps, 1, 1, 1, 1, borderType=cv2.BORDER_REFLECT)
+    batch_heatmaps_pad = np.transpose(
+        batch_heatmaps_pad.reshape((h + 2, w + 2, batch, num_kps)),
+        (2, 3, 0, 1))
+    index = coords[:, :, 0] + 1 + (coords[:, :, 1] + 1) * (w + 2)
+
+    index = index.reshape((persons, -1))
+    if share_heatmap:
```
innerlee, line 295: this if else could be rewritten as
```python
index += (w + 2) * (h + 2) * np.arange(0, N * K).reshape(-1, K)
index = index.astype(np.int).reshape(-1, 1)
```

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: 'Batch' and 'persons' are different in bottom-up paradigm

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+        offset = np.arange(0, num_kps) * (w + 2) * (h + 2)
+        index = index + offset[None, :]
+    else:
+        offset = np.arange(0, batch * num_kps) * (w + 2) * (h + 2)
+        index = index + offset.reshape((persons, -1))
+    index = index.astype(np.int).reshape((-1, 1))
+    batch_heatmaps_pad = batch_heatmaps_pad.reshape((-1, ))
```
innerlee, line 302: do you want `batch_heatmaps_pad.flatten()`?

innerlee: If so, you could flatten it at L289

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: meet the requirement of shape for the concatenate

mmpose/core/evaluation/top_down_eval.py
innerlee: Then need to update the docstring?

image

mmpose/core/evaluation/top_down_eval.py
innerlee: I guess the index already has this shape

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+    i_ = batch_heatmaps_pad[index]
```
innerlee, line 303: haven't looked at the details, could the following operations be re-written using conv?

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: yes

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: batch_heatmaps[i, j, :, :] is better

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: yes, re-written using conv may be more elegance but at the cost of more calculation?

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: yes, cv2 support this and the latter is adopted

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: done

mmpose/core/evaluation/top_down_eval.py
innerlee: why? conv should save some computations i guess

HuangJunJie2017: em, B*K*H*W (conv) vs N*K? 
batch_size: B
num_keypoints: K
num_persons: N
hight_of_heatmaps: H
width_of_heatmaps: W

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: yes, it has, and the 'reshape' operation is unnecessary

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: yes, it is more elegant and param 'share_heatmap' can be removed

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: yes, flatten() is needed

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,83 @@ def _taylor(heatmap, coord):
...
+    dyy = iy1 - 2 * i_ + iy1_
+    dxy = 0.5 * (ix1y1 - ix1 - iy1 + i_ + i_ - ix1_ - iy1_ + ix1_y1_)
+    hessian = np.concatenate([dxx, dxy, dxy, dyy], axis=1).reshape(
+        (persons, num_kps, 2, 2))
+    for i in range(persons):
+        for j in range(num_kps):
+            hessian_tmp = hessian[i, j, :, :] + \
+                          np.finfo(np.float32).eps * np.eye(2)
+            derivative_tmp = derivative[i, j, :, :]
+            shift = np.matmul(np.linalg.inv(hessian_tmp), derivative_tmp)
```
innerlee, line 325: Since it is 2x2, you can directly write the solution of inv, so that the two for loops could be avoided?

HuangJunJie2017: use this instead~
>hessian = np.linalg.inv(hessian + np.finfo(np.float32).eps * np.eye(2))
 >coords += np.einsum('ijmn,ijnk->ijmk', hessian, derivative).squeeze()

innerlee: This is better, we can do it even better. I mean, eliminate the two outer for i, j loops

HuangJunJie2017: in this way, the two outer for i,j loops have been removed

innerlee: yeah makes sense.

I'm waiting for a revision before continue :p

HuangJunJie2017: done

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,69 @@ def _taylor(heatmap, coord):
     return coord
 
 
+def post_dark(coords, batch_heatmaps, kernel=3):
+    """DARK post-pocessing.
+
+    Note:
+        batch_size: B
```
innerlee, line 257: Since this is note, `batch_size` -> `batch size`

innerlee: anyway, this is not important

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,69 @@ def _taylor(heatmap, coord):
...
+        kernel (int): Gaussian kernel size (K) for modulation.
+
+    Returns:
+        res (np.ndarray[N, K, 2]): Refined coordinates.
+    """
+    if not isinstance(batch_heatmaps, np.ndarray):
+        batch_heatmaps = batch_heatmaps.cpu().numpy()
+    B, K, H, W = batch_heatmaps.shape
+    N = coords.shape[0]
+    assert (B == 1 or B == N)
```
innerlee, line 277: superfluous brackets

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,69 @@ def _taylor(heatmap, coord):
...
+    for heatmaps in batch_heatmaps:
+        for heatmap in heatmaps:
+            cv2.GaussianBlur(heatmap, (kernel, kernel), 0, heatmap)
+    np.clip(batch_heatmaps, 0.001, 50, batch_heatmaps)
+    np.log(batch_heatmaps, batch_heatmaps)
+    batch_heatmaps = np.transpose(batch_heatmaps, (2, 3, 0, 1)).reshape(
```
innerlee, line 283: `reshape((H, W, -1))` -> `reshape(H, W, -1)`

innerlee: We prefer simpler&concise forms

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -319,6 +385,14 @@ def keypoints_from_heatmaps(heatmaps,
         kernel (int): Gaussian kernel size (K) for modulation, which should
             match the heatmap gaussian sigma when training.
             K=17 for sigma=3 and k=11 for sigma=2.
+        kpd (float): Keypoint pose distance for UDP.
```
innerlee, line 388: `kdp` is a little cryptic, how about use its full name?

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -319,6 +385,14 @@ def keypoints_from_heatmaps(heatmaps,
...
+        use_udp (bool): Use unbiased data processing.
+        target_type (str): 'GaussianHeatMap' or 'CombinedTarget'.
+            CombinedTarget: The combination of classification target
```
innerlee, line 391: Also give GaussianHeatMap a brief description

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: ok， i will update this in the next commit

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: ok， i will check this problem globally

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: Keypoint_pose_distance instead?

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: ok

mmpose/core/evaluation/top_down_eval.py
innerlee: I'm okay with long names

HuangJunJie2017: ok, point taken

mmpose/datasets/pipelines/top_down_transform.py
```diff
@@ -356,12 +382,112 @@ def _megvii_generate_target(self, cfg, joints_3d, joints_3d_visible,
...
+                if ul[0] >= heatmap_size[0] or ul[1] >= heatmap_size[1] \
+                        or br[0] < 0 or br[1] < 0:
+                    # If not, just return the image as is
+                    target_weight[joint_id] = 0
+                    continue
+
+                # # Generate gaussian
+                size = 2 * tmp_size + 1
+                x = np.arange(0, size, 1, np.float32)
+                y = x[:, np.newaxis]
```
innerlee, line 433: `np.newaxis` -> None

btw would you mind update a version for coding styles?

HuangJunJie2017: what's the meaning of 'update a version'?

innerlee: Push a commit fixing nits

HuangJunJie2017: done

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -319,6 +388,15 @@ def keypoints_from_heatmaps(heatmaps,
         kernel (int): Gaussian kernel size (K) for modulation, which should
             match the heatmap gaussian sigma when training.
             K=17 for sigma=3 and k=11 for sigma=2.
+        kps_pose_distance (float): Keypoint pose distance for UDP.
```
innerlee, line 391: `kps_pose_distance` -> `keypoint_pose_distance`

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -250,6 +250,72 @@ def _taylor(heatmap, coord):
     return coord
 
 
+def post_dark(coords, batch_heatmaps, kernel=3):
+    """DARK post-pocessing. Implemented by udp. Paper ref: Huang et al. The
```
innerlee, line 254: > Implemented by udp

Are there any sufficient differences with the original dark post-processing? If they are almost equivalent, you can remove the explanations. If they have important differences, then you can rename it to `post_dark_udp`

HuangJunJie2017: done

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: done

mmpose/core/evaluation/top_down_eval.py
innerlee: Okay I will check it tonight

innerlee: I have read the codes, the computation is okay.

The only thing to check is that this function will change (when the input is ndarray) the values of batch_heatmaps inplace, so it is kind of unsafe. Please confirm that this would not cause trouble

HuangJunJie2017: so far, it is save, as it is call at the last ~

innerlee: btw, have you tried to call this function with batch size > 1?

HuangJunJie2017: you means to use this function in top-down paradigm? yes, i committed the code after i have tested the result of three baseline top-down-GaussianHeatmap/top-down-CombinedTraget/bottom-up-GaussianHeatmap

innerlee: Another question, what if I change the padding to zero padding? Is it supposed to affect results?

HuangJunJie2017: i test this just now with two different configuration and the results show that this may have precision loss by 0.1~0.2 AP

innerlee: I mean cv2.copyMakeBorder might not support channel number > 3

image

HuangJunJie2017: emmm, it works well~ for all configs the channel number is Batch*num_keypoints>3
i use this version opencv~ 
opencv-python          3.4.2.17

innerlee: I see, it errors when bs is large

image

innerlee: Let's leave it as is for now... until someone raise an issue

HuangJunJie2017: add an assert alone with explanation to avoid this?

innerlee: Simply ignore it is okay

HuangJunJie2017: ok

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -319,6 +388,14 @@ def keypoints_from_heatmaps(heatmaps,
         kernel (int): Gaussian kernel size (K) for modulation, which should
             match the heatmap gaussian sigma when training.
             K=17 for sigma=3 and k=11 for sigma=2.
+        keypoint_pose_distance (float): Keypoint pose distance for UDP.
```
innerlee, line 391: I'm trying to understand the meaning of `keypoint pose distance`. Is it the distance of keypoint to pose?
If not, does it have more precise/descriptive name?

HuangJunJie2017: this is first proposed in google paper  ''Towards Accurate Multi-person Pose Estimation in the Wild'' where use 'radius' as denotation, and we explain this as the valid distance from the annotated keypoint coordiantes，and the pixels in the heatmap within this distance should be classify as positive in the heatmap.  'keypoint_pose_distance' is not the name i design and is inhering from other repo~ may be the original 'radius' is more reasonable?
image

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -327,35 +404,56 @@ def keypoints_from_heatmaps(heatmaps,
...
+        assert target_type in ['GaussianHeatMap', 'CombinedTarget']
+        if target_type == 'GaussianHeatMap':
+            preds, maxvals = _get_max_preds(heatmaps)
+            if post_process:
+                preds = post_dark_udp(preds, heatmaps, kernel=kernel)
+        elif target_type == 'CombinedTarget':
+            for person_heatmaps in heatmaps:
+                for i, heatmap in enumerate(person_heatmaps):
+                    kt = 2 * kernel + 1 if i % 3 == 0 else kernel
+                    cv2.GaussianBlur(heatmap, (kt, kt), 0, heatmap)
```
innerlee, line 418: The heatmaps are smoothed inplace. Again, double check that it is safe.

Actually I want to suggest to copy the  heatmaps before this process, so that the function is polite and safe

mmpose/core/evaluation/top_down_eval.py
innerlee: Thanks for the clear explanation. How about `valid_radius`? And in the docstring, give it concreate descriptions like `The predicted positions within a disk of radius R from the location of a keypoint is considered as correct prediction to that keypoint`.

mmpose/core/post_processing/group.py
```diff
@@ -300,6 +304,8 @@ def refine(heatmap, tag, keypoints):
             if keypoints[i, 2] > 0:
                 # save tag value of detected keypoint
                 x, y = keypoints[i][:2].astype(int)
+                x = max(0, min(x, W - 1))
```
innerlee, line 307: np.clip

mmpose/core/post_processing/post_transforms.py
```diff
@@ -62,26 +62,34 @@ def flip_back(output_flipped, flip_pairs):
             from the flipped images.
         flip_pairs (list[tuple()): Pairs of keypoints which are mirrored
             (for example, left ear -- right ear).
+        target_type (str): GaussianHeatMap or CombinedTarget
 
     Returns:
         np.ndarray: heatmaps that flipped back to the original image
     """
     assert output_flipped.ndim == 4, \
         'output_flipped should be [batch_size, num_keypoints, height, width]'
```
innerlee, line 72: add `assert target_type in ('GaussianHeatMap', 'CombinedTarget')`

mmpose/core/post_processing/post_transforms.py
```diff
@@ -112,11 +121,20 @@ def transform_preds(coords, center, scale, output_size):
     assert len(center) == 2
     assert len(scale) == 2
     assert len(output_size) == 2
-
-    target_coords = coords.copy()
-    trans = get_affine_transform(center, scale, 0, output_size, inv=True)
-    for p in range(coords.shape[0]):
-        target_coords[p, 0:2] = affine_transform(coords[p, 0:2], trans)
+    if use_udp:
+        scale = scale * 200.0
```
innerlee, line 125: add a comment on the magic number 200

mmpose/datasets/pipelines/bottom_up_transform.py
```diff
@@ -6,19 +8,68 @@
...
+        theta (float): Rotation angle in degrees.
+        size_input (np.ndarray): Size of input image [w, h].
+        size_dst (np.ndarray): Size of output image [w, h].
+        size_target (np.ndarray): Size of ROI in input plane [w, h].
+
+    Returns:
+        matrix (np.ndarray): A matrix for transformation.
+    """
+    size_target = size_target
+    theta = theta / 180.0 * math.pi
```
innerlee, line 26: there is a `np.deg2rad`

mmpose/datasets/pipelines/bottom_up_transform.py
```diff
@@ -6,19 +8,68 @@
...
+    Args:
+        theta (float): Rotation angle in degrees.
+        size_input (np.ndarray): Size of input image [w, h].
+        size_dst (np.ndarray): Size of output image [w, h].
+        size_target (np.ndarray): Size of ROI in input plane [w, h].
+
+    Returns:
+        matrix (np.ndarray): A matrix for transformation.
+    """
+    size_target = size_target
```
innerlee, line 25: What does this line do

mmpose/datasets/pipelines/bottom_up_transform.py
```diff
@@ -6,19 +8,68 @@
 from .shared_transform import Compose
 
 
+def get_warpmatrix(theta, size_input, size_dst, size_target):
```
innerlee, line 11: `get_warp_matrix`

mmpose/datasets/pipelines/bottom_up_transform.py
```diff
@@ -6,19 +8,68 @@
...
+                              0.5 * size_target[0])
+    matrix[1, 0] = math.sin(theta) * scale_y
+    matrix[1, 1] = math.cos(theta) * scale_y
+    matrix[1, 2] = scale_y * (-0.5 * size_input[0] * math.sin(theta) -
+                              0.5 * size_input[1] * math.cos(theta) +
+                              0.5 * size_target[1])
+    return matrix
+
+
+def affine_joints(joints, mat):
```
innerlee, line 43: affine is an adjective

mmpose/datasets/pipelines/bottom_up_transform.py
```diff
@@ -6,19 +8,68 @@
...
+    """Affine the joints by the transform matrix."""
```
innerlee, line 44: `Apply affine transformation defined by the transformation matrix on joints`

innerlee: What's the typical shape of `joints`?

mmpose/datasets/pipelines/bottom_up_transform.py
```diff
@@ -6,19 +8,68 @@
...
+    joints = np.array(joints)
+    shape = joints.shape
+    joints = joints.reshape(-1, 2)
+    return np.dot(
+        np.concatenate((joints, joints[:, 0:1] * 0 + 1), axis=1),
```
innerlee, line 49: `joints[:, 0:1] * 0 + 1` was used to create an all one matrix?

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: 'valid_radius' is better for understanding.  i prefer to explain this as 'The radius of the positive area in classification heatmap.'

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: it is save here

mmpose/core/post_processing/group.py
HuangJunJie2017: done

mmpose/datasets/pipelines/top_down_transform.py
```diff
@@ -179,16 +188,24 @@ def __call__(self, results):
...
-            trans, (int(image_size[0]), int(image_size[1])),
-            flags=cv2.INTER_LINEAR)
 
-        for i in range(results['ann_info']['num_joints']):
-            if joints_3d_visible[i, 0] > 0.0:
-                joints_3d[i, 0:2] = affine_transform(joints_3d[i, 0:2], trans)
+        if self.use_udp:
+            trans = get_warpmatrix(r, c * 2.0, image_size - 1.0, s * 200.0)
+            img = cv2.warpAffine(
+                img.copy(),
```
innerlee, line 195: is this copy needed?

mmpose/core/post_processing/post_transforms.py
HuangJunJie2017: done

mmpose/datasets/pipelines/bottom_up_transform.py
innerlee: How about the name `transform_joints_affine`?

mmpose/models/detectors/top_down.py
```diff
@@ -153,26 +154,27 @@ def forward_train(self, img, target, target_weight, img_metas, **kwargs):
...
+        if self.target_type == 'GaussianHeatMap':
+            if isinstance(output, list):
+                if target.dim() == 5 and target_weight.dim() == 4:
+                    _, avg_acc, _ = pose_pck_accuracy(
+                        output[-1].detach().cpu().numpy(),
+                        target[:, -1, ...].detach().cpu().numpy(),
+                        target_weight[:, -1,
+                                      ...].detach().cpu().numpy().squeeze(-1) >
+                        0)
+                    # Only use the last output for prediction
```
innerlee, line 166: where does this line of comment applies to, the if part, or the else part?

mmpose/core/post_processing/post_transforms.py
HuangJunJie2017: 'inverse the effect of scale normalize'?
'the scale is normalize by factor of 200.0 in 'mmpose/datasets/datasets/top_down/topdown_coco_dataset.py,line217-244''

mmpose/datasets/pipelines/bottom_up_transform.py
HuangJunJie2017: got it, done

mmpose/models/losses/mse_loss.py
```diff
@@ -42,6 +42,53 @@ def forward(self, output, target, target_weight):
         return loss / num_joints
 
 
+@LOSSES.register_module()
+class JointsMSELoss_Combined(nn.Module):
```
innerlee, line 46: name candidate: `CombinedJointsMSELoss`

innerlee:
Generally in good shape. Looks good after some style changes.

The next steps before merge:

* add unittests
* train checkpoints (cc. @jin-s13)

mmpose/core/evaluation/top_down_eval.py
innerlee: Yeah that's better

mmpose/datasets/pipelines/bottom_up_transform.py
HuangJunJie2017: it is meaning less now, i will check and remove this

mmpose/datasets/pipelines/bottom_up_transform.py
HuangJunJie2017: this name is inherit from the origin mmpose project https://github.com/open-mmlab/mmpose/blob/master/mmpose/datasets/pipelines/bottom_up_transform.py in line260

mmpose/datasets/pipelines/bottom_up_transform.py
HuangJunJie2017: yes

HuangJunJie2017: this function is inherit from the origin mmpose project https://github.com/open-mmlab/mmpose/blob/master/mmpose/datasets/pipelines/bottom_up_transform.py in line260

mmpose/datasets/pipelines/bottom_up_transform.py
innerlee: It was private function before, so the naming was more or less arbitrary.

Now it does not have `_` prefix

mmpose/models/losses/mse_loss.py
HuangJunJie2017: JointsMSELoss_CombinedTarget?

mmpose/models/detectors/top_down.py
HuangJunJie2017: this comment is inherit from the origin mmpose project

mmpose/models/losses/mse_loss.py
innerlee: Try again with

* avoid `_`
* end with `Loss`

mmpose/datasets/pipelines/bottom_up_transform.py
HuangJunJie2017: point taken, we can imitate the function name cv2.warpAffine(),  we have warpAffineJoints/warpAffineKeypoints/warpAffine_Joints/warpAffine_Keypoints/?

mmpose/models/detectors/top_down.py
innerlee: I see, it should be placed before L163

mmpose/datasets/pipelines/bottom_up_transform.py
innerlee: `warp_affine_joints`

mmpose/models/losses/mse_loss.py
HuangJunJie2017: CombinedTargetJointsMSELoss? CombinedJoints is kind of strange, as 'Combined' is use to describe the targets instead of the joints

innerlee: `CombinedTargetMSELoss`?

mmpose/datasets/pipelines/bottom_up_transform.py
HuangJunJie2017: 'warp_affine_joints' is adopted

mmpose/models/losses/mse_loss.py
HuangJunJie2017: 'CombinedTargetMSELoss' is better, 'Joints' in 'CombinedTargetJointsMSELoss' is quite unnecessary.

innerlee: Great

mmpose/datasets/pipelines/top_down_transform.py
HuangJunJie2017: i check this, it is not necessary, i will remove this in both top-down and bottom-up transformation

innerlee:
@HuangJunJie2017 About unittests, what's your opinion? Would you like to add them by yourself, or you prefer to handle the task to us?

HuangJunJie2017:
> @HuangJunJie2017 About unittests, what's your opinion? Would you like to add them by yourself, or you prefer to handle the task to us?

emm, sincerely i don' t know what is unittests. can you give me a brief introduction or an example?

HuangJunJie2017:
the assertion checking design in  mmpose/test/ ? i think i can try to design this~

innerlee:
Unittesting is a way to make sure that codes are usable, and correct. This is done by running a set of functions in the `tests/` folder. The `unit` in unittesting means that every single function is checked individually.
For example, these lines https://github.com/open-mmlab/mmpose/blob/master/tests/test_pipelines/test_top_down_pipelines.py#L123-L126 check if `TopDownRandomFlip` is working properly.

To run tests, type the following command in the root folder of mmpose
```bash
pytest
```

innerlee:
Generally speaking, every newly added functions/objects need to be tested. Which function is not tested yet is listed in the coverage report, see https://codecov.io/gh/open-mmlab/mmpose/pull/291/diff?src=pr&el=tree-more

HuangJunJie2017:
i see, i prefer to handle this task to you~
> https://codecov.io/gh/open-mmlab/mmpose/pull/291/diff?src=pr&el=tree-more
this is an address of intranet~

innerlee:
That's fine. Two last things from your side:

1. resolve the code conflicts, and
2. provide a more complete set of configs, so that we can train checkpoints

💟

mmpose/core/evaluation/top_down_eval.py
```diff
@@ -349,6 +427,8 @@ def keypoints_from_heatmaps(heatmaps,
         assert post_process not in [False, None, 'megvii']
     if post_process in ['megvii', 'unbiased']:
         assert kernel > 0
+    if use_udp:
```
innerlee, line 430: Looking at the change, I'm wandering if we can add an `udp` option in the `post_process` argument? What do you think?

innerlee:
Sorry forgot an important thing. Each algorithm has its own README file, in which the paper information is provided. Here is an example for darkpose https://github.com/open-mmlab/mmpose/tree/master/configs/top_down/darkpose

UDP surely deserves a separate folder in configs.

How about put all top down udp configs under `configs/top_down/udp`, and bottom up configs at `configs/bottom_up/udp`?

mmpose/core/evaluation/top_down_eval.py
HuangJunJie2017: emmm, i think it is unnecessary, as i have remove the concept of 'post_process' in udp~  i add an assertion here to prevent this:
>if post_process == 'megvii':
>        maxvals = maxvals / 255.0 + 0.5

HuangJunJie2017:
https://github.com/open-mmlab/mmpose/pull/291#discussion_r528499791
we discuss above and this is the plan i use now

HuangJunJie2017:
four example configs have been pushed
btw if need, i can provide the train log and pretrained models for reference~

innerlee (approved):

innerlee:
Thanks! Please post the performances here for future reference.

I plan to merge this tomorrow morning. Unittests and checkpoints will be done in separate prs.

HuangJunJie2017:
logs of train & test is provided here
https://drive.google.com/drive/folders/17nP0-OaF372ipBtTZ6hTTcCmRVCTWWwn

PR #294, opened 2020-11-22
pylint remove object from bases

### LINKED ISSUE #250: Pylint: R0205
```bash
mmpose/datasets/pipelines/mesh_transform.py:169:0: R0205: Class 'LoadIUVFromFile' inherits from object, can be safely removed from bases in python3 (useless-object-inheritance)
```

jin-s13:
fix #250

### DIFF
--- a/mmpose/datasets/pipelines/mesh_transform.py
+++ b/mmpose/datasets/pipelines/mesh_transform.py
@@ -167,7 +167,7 @@ def _flip_joints_3d(joints_3d, joints_3d_visible, flip_pairs):
 
 
 @PIPELINES.register_module()
-class LoadIUVFromFile(object):
+class LoadIUVFromFile():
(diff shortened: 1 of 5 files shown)


mmpose/datasets/pipelines/mesh_transform.py
```diff
@@ -167,7 +167,7 @@ def _flip_joints_3d(joints_3d, joints_3d_visible, flip_pairs):
 
 
 @PIPELINES.register_module()
-class LoadIUVFromFile(object):
+class LoadIUVFromFile():
```
innerlee, line 170: the `()` could be removed also

jin-s13: Done.

innerlee (approved):

PR #296, opened 2020-11-22
Human mesh readme

### LINKED ISSUE #179: Data preparation for H36M
the preparation docs for H36M

jin-s13:
fix #179


docs/README_human_mesh.md
```diff
@@ -0,0 +1,294 @@
+# Human Mesh Estimation
```
innerlee, line 1: The filename does not look right.
Where did we put those files before, `data_preparation.md`?

docs/data_preparation.md
```diff
@@ -16,10 +16,15 @@ MMPose supported datasets:
 - [FreiHand](https://lmb.informatik.uni-freiburg.de/projects/freihand/)
 - [CMU Panoptic HandDB](http://domedb.perception.cs.cmu.edu/handdb.html)
 - [InterHand2.6M](https://mks0601.github.io/InterHand2.6M/)
+- [Human3.6M](http://vision.imar.ro/human3.6m/description.php)
+- [MPI-INF-3DHP](http://gvv.mpi-inf.mpg.de/3dhp-dataset/)
```
innerlee, line 20: Which dataset type should these new datasets use?

innerlee: You see, we have many specific dataset classes

image

Which class should be used for those new datasets?

jin-s13: These datasets are used in `MeshMixDataset` see   https://github.com/open-mmlab/mmpose/blob/f456debefab5d8b29bbf62785f76077a0b9b4a12/configs/mesh/hmr/hmr_resnet_50.py#L116

jin-s13: When training HMR, many datasets are used, including:
h36m, mpi_inf_3dhp, lsp, lspet, mpii, coco2014

docs/README_human_mesh.md
jin-s13: Since this is a different task (human mesh recovery), maybe we should use a separate README file?
Need further discussion.

innerlee: I see, this is a doc for tasks. Previously supported tasks haven't got their readmes yet.
How about put this file at `configs/mesh/README.md`?
Later we can add `configs/hand/README.md`, etc.

innerlee: We can use script to gather them into the docs site

jin-s13: agree

docs/data_preparation.md
innerlee: okay

configs/mesh/README.md
```diff
@@ -0,0 +1,294 @@
+# Human Mesh Estimation
+
```
innerlee, line 2: add a brief intro to this task here

innerlee (approved):

PR #318, opened 2020-11-27
add use_nms option

wusize:
Add use_nms option

### DIFF
--- a/mmpose/datasets/datasets/top_down/topdown_aic_dataset.py
+++ b/mmpose/datasets/datasets/top_down/topdown_aic_dataset.py
@@ -54,11 +54,7 @@ class TopDownAicDataset(TopDownCocoDataset):
         self.use_gt_bbox = data_cfg['use_gt_bbox']
         self.bbox_file = data_cfg['bbox_file']
         self.image_thr = data_cfg['image_thr']
-
-        if 'use_nms' in data_cfg:
-            self.use_nms = data_cfg['use_nms']
-        else:
-            self.use_nms = True
+        self.use_nms = data_cfg.get('use_nms', True)
         self.soft_nms = data_cfg['soft_nms']
         self.nms_thr = data_cfg['nms_thr']
         self.oks_thr = data_cfg['oks_thr']
--- a/mmpose/datasets/datasets/top_down/topdown_aic_dataset.py
+++ b/mmpose/datasets/datasets/top_down/topdown_aic_dataset.py
@@ -55,6 +55,10 @@ class TopDownAicDataset(TopDownCocoDataset):
         self.bbox_file = data_cfg['bbox_file']
         self.image_thr = data_cfg['image_thr']
 
+        if 'use_nms' in data_cfg:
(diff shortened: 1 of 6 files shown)


mmpose/datasets/datasets/top_down/topdown_aic_dataset.py
```diff
@@ -55,6 +55,10 @@ def __init__(self,
         self.bbox_file = data_cfg['bbox_file']
         self.image_thr = data_cfg['image_thr']
 
+        if 'use_nms' in data_cfg:
```
innerlee, line 58: `data_cfg.get('use_nms', True)

jin-s13: fixed.

mmpose/datasets/datasets/top_down/topdown_coco_dataset.py
```diff
@@ -361,15 +361,18 @@ def evaluate(self, outputs, res_folder, metric='mAP', **kwargs):
                 # rescoring
                 n_p['score'] = kpt_score * box_score
 
-            nms = soft_oks_nms if self.soft_nms else oks_nms
-            keep = nms(list(img_kpts), oks_thr, sigmas=self.sigmas)
+            if self.use_nms:
```
innerlee, line 364: `if self.use_nms and len(img_kpts) > 0:`

innerlee: According to discussion, nms will always return values if input is non-empty

wusize: We might need these lines in case of len(img_kpts)=0.

innerlee: see the code posted above

wusize: Em,,, I mean it's very likely that **"img_kpts = kpts[image_id]"** gets an empty list [ ], in the case of which nms will return an empty list [ ] and **"if len(keep) == 0: valid_kpts.append(img_kpts)"** would be necessary. 
2020-11-29 23-47-14屏幕截图

innerlee: Okay, so nms does not guarantee that the return be non-empty when img_kpts is non-empty?

wusize: Oh, now I get the point. I erroneously thought that **[img_kpts[_keep] for _keep in keep]** would raise raise if **keep=[]** and **img_kpts=[]**, but it won't.
2020-11-30 08-28-08屏幕截图
And sure nms does guarantee that. I was talking about the empty case but now it's also not a problem. So **if len(keep) == 0: valid_kpts.append(img_kpts)** is unnecessary in both cases. I will rm these lines.

innerlee (approved):

PR #328, opened 2020-12-01
fix bugs in pck calculation

jin-s13:


mmpose/core/evaluation/top_down_eval.py
```diff
@@ -16,17 +16,20 @@ def _calc_distances(preds, targets, mask, normalize):
     Args:
         preds (np.ndarray[N, K, 2]): Predicted keypoint location.
         targets (np.ndarray[N, K, 2]): Groundtruth keypoint location.
-        normalize (np.ndarray[N, 2]): Typical value is heatmap_size/10
```
innerlee, line 19: what happened to this `/10`

jin-s13: `/10` or `0.1 ` should not be in `normalize`, but `thr`.
In other words, `thr=0.1` and `normalize=heatmap_size`, is more reasonable.

innerlee (approved):

innerlee:
Add a test case that making the previous code error?

jin-s13:
Will add a test code soon.
And some results in README, should be updated.

PR #345, opened 2020-12-10
Add information dropping augmentation (AID)

HuangJunJie2017:
Add information dropping augmentation (AID)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,10 +35,10 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7]
-        torch: [1.3.1, 1.5.1, 1.6.0, 1.7.0]
+        torch: [1.3.0, 1.5.1, 1.6.0, 1.7.0]
         include:
-          - torch: 1.3.1
-            torchvision: 0.4.2
+          - torch: 1.3.0
+            torchvision: 0.4.1
           - torch: 1.5.1
             torchvision: 0.6.1
           - torch: 1.6.0
@@ -77,11 +77,11 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7]
-        torch: [1.3.1, 1.5.1+cu101, 1.6.0+cu101, 1.7.0+cu101]
+        torch: [1.3.0, 1.5.1+cu101, 1.6.0+cu101, 1.7.0+cu101]
         include:
-          - torch: 1.3.1
-            torchvision: 0.4.2
-            mmcv: "latest+torch1.3.1+cu101"
+          - torch: 1.3.0
+            torchvision: 0.4.1
+            mmcv: "latest+torch1.3.0+cu101"
           - torch: 1.5.1+cu101
             torchvision: 0.6.1+cu101
             mmcv: "latest+torch1.5.1+cu101"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,7 +58,7 @@ jobs:
       - name: Install PyTorch
         run: pip install torch==${{matrix.torch}}+cpu torchvision==${{matrix.torchvision}}+cpu -f https://download.pytorch.org/whl/torch_stable.html
       - name: Install MMCV
-        run: pip install mmcv-full==latest+torch${{matrix.torch}}+cpu -f https://download.openmmlab.com/mmcv/dist/index.html
+        run: pip install mmcv-full==1.2.2 -f https://download.openmmlab.com/mmcv/dist/cpu/torch${{matrix.torch}}/index.html
       - name: Install unittest dependencies
         run: pip install -r requirements.txt
       - name: Build and install
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,23 +83,23 @@ jobs:
         include:
           - torch: 1.3.0
             torchvision: 0.4.1
-            mmcv: "latest+torch1.3.0+cu101"
+            mmcv: "1.2.2+torch1.3.0+cu101"
           - torch: 1.5.1+cu101
             torchvision: 0.6.1+cu101
-            mmcv: "latest+torch1.5.1+cu101"
+            mmcv: "1.2.2+torch1.5.1+cu101"
           - torch: 1.6.0+cu101
             torchvision: 0.7.0+cu101
-            mmcv: "latest+torch1.6.0+cu101"
+            mmcv: "1.2.2+torch1.6.0+cu101"
           - torch: 1.7.0+cu101
             torchvision: 0.8.1+cu101
-            mmcv: "latest+torch1.7.0+cu101"
+            mmcv: "1.2.2+torch1.7.0+cu101"
           - torch: 1.7.0+cu101
             torchvision: 0.8.1+cu101
-            mmcv: "latest+torch1.7.0+cu101"
+            mmcv: "1.2.2+torch1.7.0+cu101"
             python-version: 3.6
           - torch: 1.7.0+cu101
             torchvision: 0.8.1+cu101
-            mmcv: "latest+torch1.7.0+cu101"
+            mmcv: "1.2.2+torch1.7.0+cu101"
             python-version: 3.8
 
     steps:
@@ -108,6 +108,8 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Upgrade pip
+        run: pip install pip --upgrade
       - name: Install CUDA
         run: |
           export INSTALLER=cuda-repo-${UBUNTU_VERSION}_${CUDA}_amd64.deb
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7]
-        torch: [1.3.0, 1.5.0, 1.6.0, 1.7.0]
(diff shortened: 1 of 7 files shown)


jin-s13:
Thanks for the contribution!
AID is a good collection of multiple augmentation techniques. I believe it would be better (more clear) to divide the single AID pipeline into several sub-pipelines, namely hide_and_seek, cutout and so on. What do you think? @HuangJunJie2017 
I also suggest adding config directories named AID (mmpose/configs/top_down/AID and mmpose/configs/bottom_up/AID) to collect all the related config files. Also a corresponding readme.md to cite AID.

HuangJunJie2017:
@jin-s13  divided implement and adding config directories named AID (mmpose/configs/top_down/AID and mmpose/configs/bottom_up/AID)  is ok if is the final decision, you can make discussion with other reviewers and make a final decision. And i will modify it according to the final decision.

innerlee:
Yeah implementing the component augmentations as individual, independent, composable pipelines would be helpful to others too.

The folder name is using lower case, namely `mmpose/configs/top_down/aid`

HuangJunJie2017:
point taken,
i will split the class AID into  two classes : Cutout and Hide_and_seek
beside, the configs for AID will be placed in mmpose/configs/top_down/aid/
btw,  should the configs for udp be placed in mmpose/configs/top_down/udp/ ?

jin-s13:
We have already moved these config files for udp to a separate directory, namely `mmpose/configs/top_down/udp/` : )

configs/bottom_up/aid/coco/higher_hrnet32_coco_512x512_udp_aid.py
```diff
@@ -136,7 +136,7 @@
         trans_factor=40,
         use_udp=True),
     dict(type='BottomUpRandomFlip', flip_prob=0.5),
-    dict(type='AID', prob_has=1.0),
+    dict(type='Hide_and_seek'),
```
innerlee, line 139: `Hide_and_seek` -> `HideAndSeek`

HuangJunJie2017: ok

HuangJunJie2017:
image
Need i  fix this check failures?  is that means adding unittest is required?

jin-s13:
> image
> Need i fix this check failures? is that means adding unittest is required?

Yes, unittest is needed. Please add unittests for Cutout and HideAndSeek in `tests/test_pipelines/test_top_down_pipelines.py`.

innerlee:
Yeah unittests are important and are required.

Current testing files can be used as implementation reference.

HuangJunJie2017:
ok, thanks for your patient, i will try to design this by imitating the existing unittests.

HuangJunJie2017:
when pytest locally, i run into this error, it seem an import error
image

innerlee:
It seems that the `smplx` package met some problem

HuangJunJie2017:
emm, another new error 
image

innerlee:
😕 

Try to modify this line to 1.5.1?

https://github.com/open-mmlab/mmpose/blob/master/.github/workflows/build.yml#L42

HuangJunJie2017:
is that means there are some problems in the unittest i add to this project?

innerlee:
No, not this specific one

innerlee:
Search & replace all `1.5.0`?

HuangJunJie2017:
emm, the same problem is happen in torch 1.3.0~

innerlee:
Well, let's change it again

image

innerlee:
Now mmcv errors, try to add `pip install pip --upgrade` in the begining

innerlee:
@HuangJunJie2017 CI issue has been resolved. There was `pip` problem and `mmcv` problems recently.

HuangJunJie2017:
great!

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -140,3 +141,94 @@ def __repr__(self):
...
+                 prob_has=1.0,
+                 prob_has_hide=0.5,
+                 grid_sizes=(0, 16, 32, 44, 56)):
+        self.prob_has = prob_has
+        self.prob_has_hide = prob_has_hide
+        self.grid_sizes = grid_sizes
+
+    def _hide_and_seek(self, img):
+        # get width and height of the image
+        ht, wd, _ = img.shape
```
jin-s13, line 168: height, width, _ = img.shape

HuangJunJie2017: done

jin-s13 (approved):

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -140,3 +141,94 @@ def __repr__(self):
         """Compute the string representation."""
         return (f'{self.__class__.__name__}('
                 f'keys={self.keys}, meta_keys={self.meta_keys})')
+
+
+@PIPELINES.register_module()
+class HideAndSeek:
+    """Augmentation by informantion dropping in Hide-and-Seek paradigm. Paper
```
innerlee, line 148: To help future readers and users, please provide a desription on difference between this implementation and the original paper's version, and also add reference to the original paper.

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -140,3 +141,94 @@ def __repr__(self):
...
+        img = results['img']
+        if np.random.rand() < self.prob_has:
+            img = self._hide_and_seek(img)
+        results['img'] = img
+        return results
+
+
+@PIPELINES.register_module()
+class Cutout:
+    """Augmentation by informantion dropping in Cutout paradigm. Paper ref:
```
innerlee, line 194: To help future readers and users, please provide a desription on difference between this implementation and the original paper's version, and also add reference to the original paper.

innerlee:
Sorry for the late reply. Some minor comments

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -140,3 +141,94 @@ def __repr__(self):
...
+    Estimation with Information Dropping Augmentation (arXiv:2008.07139 2020).
+
+    Args:
+        prob_has (float): Probability of performing hide-and-seek.
+        prob_has_hide (float): Probability of hiding patches.
+        grid_sizes (list): List of optional grid sizes.
+    """
+
+    def __init__(self,
+                 prob_has=1.0,
```
innerlee, line 159: `prob_has` -> `prob`?

innerlee: or `prob_hide_and_seek`?

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -140,3 +141,94 @@ def __repr__(self):
...
+                 prob_has_hide=0.5,
```
innerlee, line 160: `prob_has_hide` -> `prob_hiding_patches`?

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -140,3 +141,94 @@ def __repr__(self):
...
+        feat_x_int, feat_y_int = np.meshgrid(feat_x_int, feat_y_int)
+        feat_x_int = feat_x_int.flatten()
+        feat_y_int = feat_y_int.flatten()
+        for _ in range(self.num_patch):
+            center = [np.random.rand() * width, np.random.rand() * height]
+            radius = self.radius_factor * (1 + np.random.rand(2)) * width
+            x_offset = (center[0] - feat_x_int) / radius[0]
+            y_offset = (center[1] - feat_y_int) / radius[1]
+            dis = x_offset**2 + y_offset**2
+            indexes = np.where(dis <= 1)[0]
```
innerlee, line 224: Havn't acturally tried this code. It looks inefficient. Just guessing. Have you experienced slowness on training?

HuangJunJie2017: I tried this, and the latency in training is negligible

innerlee: Okay

mmpose/datasets/pipelines/shared_transform.py
HuangJunJie2017: I think it is unnecessary and improper to stated this here. As this should be found in Paper ref: Huang et al. AID: Pushing the Performance Boundary of Human Pose Estimation with Information Dropping Augmentation (arXiv:2008.07139 2020). This paper has detailed the relationship.

mmpose/datasets/pipelines/shared_transform.py
HuangJunJie2017: ok， prob is enough

mmpose/datasets/pipelines/shared_transform.py
HuangJunJie2017: 'prob_hiding_patches' is better

.github/workflows/build.yml
```diff
@@ -35,11 +35,11 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7]
-        torch: [1.3.0, 1.5.0, 1.6.0, 1.7.0]
+        torch: [1.3.0, 1.5.1, 1.6.0, 1.7.0]
```
innerlee, line 38: Changes of this file could be dropped, since CI has been fixed.

HuangJunJie2017: emmm, why there is no commit conflict?

PR #357, opened 2020-12-16
update faq

### LINKED ISSUE #356: About unexpected keys for hrnet
Hello, I'm new to mmpose.
When I train hrnetw32 with the default config. I get this warning message. I print the state_dict keys to debug, and fail to figure out why. Could you kindly lend me a hand here?
image

jin-s13:
fix #356

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -46,6 +46,10 @@ If the contents here do not cover your issue, please create an issue using the [
     Set the environment variables `MASTER_PORT=XXX`. For example,
     `MASTER_PORT=29517 GPUS=16 GPUS_PER_NODE=8 CPUS_PER_TASK=2 ./tools/slurm_train.sh Test res50 configs/top_down/resnet/coco/res50_coco_256x192.py work_dirs/res50_coco_256x192`
 
+- **"Unexpected keys in source state dict" when loading pre-trained weights**
+
+    That's normal. ImageNet-pretrained classification network and the pose network may have different architectures (e.g. classification head).
+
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -46,6 +46,10 @@ If the contents here do not cover your issue, please create an issue using the [
     Set the environment variables `MASTER_PORT=XXX`. For example,
     `MASTER_PORT=29517 GPUS=16 GPUS_PER_NODE=8 CPUS_PER_TASK=2 ./tools/slurm_train.sh Test res50 configs/top_down/resnet/coco/res50_coco_256x192.py work_dirs/res50_coco_256x192`
 
+- **"Unexpected keys in source state dict" when loading pre-trained weights**
+
+    That's normal. ImageNet-pretrained classification network and the pose network may have different architectures (e.g. classification head).
```
innerlee, line 51: It's normal that some layers in the pretrained model are not used in the pose model. ImageNet-pretrained classification network and the pose network may have different architectures (e.g. no classification head). So some `unexpected keys in source state dict` is actually expected.

innerlee (approved):

docs/faq.md
qiuzhongwei-USTB: But some keys in the backbone are still unexpected! like stage4.2.fuse_layers.3.2.0.1.running_mean, 
stage4.2.fuse_layers.3.2.0.1.running_var, stage4.2.fuse_layers.3.2.0.1.num_batches_tracked ...
They are not head. So it's also normal?

image

PR #370, opened 2020-12-18
add resnest and vgg backbone

jin-s13:
The codes are from MMClassification.


innerlee:
Did you run some configs containing them, for the sake of sanity check?

jin-s13:
Not yet. Will add some related config files and checkpoints later.

innerlee:
Also update repo's README

innerlee (approved):

PR #380, opened 2020-12-24
add pose tracking demo

luminxu:
Get track id using greedy bbox IOU tracking & Visualize persons according to track id.


innerlee:
Need unittests and README like https://github.com/open-mmlab/mmpose/blob/master/demo/2d_hand_demo.md

demo/2d_pose_tracking_demo.md
```diff
@@ -0,0 +1,37 @@
+## 2D Pose Tracking Demo
```
innerlee, line 1: add this file into demo/README.md

mmpose/apis/inference_tracking.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+
+    xA = max(boxA[0], boxB[0])
+    yA = max(boxA[1], boxB[1])
+    xB = min(boxA[2], boxB[2])
+    yB = min(boxA[3], boxB[3])
+
+    interArea = max(0, xB - xA + 1) * max(0, yB - yA + 1)
+
+    boxAArea = (boxA[2] - boxA[0] + 1) * (boxA[3] - boxA[1] + 1)
```
jin-s13, line 22: boxA_area

mmpose/apis/inference_tracking.py
```diff
@@ -0,0 +1,196 @@
...
+    Returns:
+        float: The IoU value.
+    """
+
+    xA = max(boxA[0], boxB[0])
+    yA = max(boxA[1], boxB[1])
+    xB = min(boxA[2], boxB[2])
+    yB = min(boxA[3], boxB[3])
+
+    interArea = max(0, xB - xA + 1) * max(0, yB - yA + 1)
```
jin-s13, line 20: inter_area

mmpose/apis/inference_tracking.py
```diff
@@ -0,0 +1,196 @@
...
+    yA = max(boxA[1], boxB[1])
+    xB = min(boxA[2], boxB[2])
+    yB = min(boxA[3], boxB[3])
+
+    inter_area = max(0, xB - xA + 1) * max(0, yB - yA + 1)
+
+    boxA_area = (boxA[2] - boxA[0] + 1) * (boxA[3] - boxA[1] + 1)
+    boxB_area = (boxB[2] - boxB[0] + 1) * (boxB[3] - boxB[1] + 1)
+
+    iou = inter_area / float(boxA_area + boxB_area - inter_area)
```
innerlee, line 25: Any chance that the denominator equals 0?

innerlee: Will this function be used elsewhere?

mmpose/apis/inference_tracking.py
```diff
@@ -0,0 +1,196 @@
...
+def get_track_id(results, results_last, next_id, iou_thresh=0.3):
+    """Get track id for each person instance on the current frame.
+
+    Args:
+        results (list[dict]): The bbox & pose results of the current frame
+                (bbox_result, pose_result).
+        results_last (list[dict]): The bbox & pose & track_id info of the
+                last frame (bbox_result, pose_result, track_id).
+        next_id (int): The track id for the new person instance.
+        iou_thresh (float): The threshold for iou tracking.
```
innerlee, line 76: thresh? or thr?

mmpose/apis/inference_tracking.py
```diff
@@ -0,0 +1,196 @@
...
+
+    Returns:
+        list[dict]: The bbox & pose & track_id info of the
+                current frame (bbox_result, pose_result, track_id).
+        int: The track id for the new person instance.
+    """
+
+    for i in range(len(results)):
```
innerlee, line 84: `for result in results:`

mmpose/apis/inference_tracking.py
luminxu: Do not worry as boxA_area >= inter_area & boxB_area >= inter_area. The denominator equals 0 only is boxA_area = boxA_area = 0, which is impossible for mmdet to predict such a person bbox.

jin-s13 (approved):

jin-s13:
The unittest fails.

jin-s13 (changes requested):
The unittest fails.

mmpose/apis/inference_tracking.py
jin-s13: Do we need assertion that `boxA[2] > boxA[0]` and `boxA[3] > boxA[1]` ? @innerlee

jin-s13 (approved):

mmpose/apis/inference_tracking.py
innerlee: If a zero-area bbox be considered an error, we can add assertions.
If it is allowed, then we may add a `max(1, blablha...)` at the denominator.

BTW what's the difference between `box` and `bbox`?

luminxu: I think adding assertions should be better. And one suggestion is that the previous steps may also consider this kind of error (box[2] <= box[0]). In detail, if such a bbox is given, function inference_top_down_pose_model may have some problems. We may also consider to add assertions in function _xyxy2xywh or anywhere suitable.

As for the difference box and bbox, function _compute_iou is more general and can compute iou for any two boxes. I use box because I do not want to emphasize they are person bounding boxes. However, we may not consider other kinds of boxes and I will use bbox to avoid confusion.

If you gays agree with my opinions, I will make these two changes very soon. @innerlee @jin-s13

innerlee: More checks are good. One thing to take care is that we 
* use assertion only when the it is an error, because a violation would stop the running program, and any unsaved computations would got lost. 
* If the case is not critical, then a warning is good enough. 
* If it is neglectable, we can simply ignore it and replace with safe neutral values

I vote for bbox.

OpenMMLab-gd:
Dear luminxu,
First of all, we want to express our gratitude for your significant PR in the MMPose project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.
If you are Chinese or have WeChat，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）
We would also like to invite you to join our Special Interest Group (SIG) private channel on Discord, where you can share your experiences, ideas, and build connections with like-minded peers. To join the SIG channel, simply message moderator— OpenMMLab on Discord or briefly share your open-source contributions in the #introductions channel and we will assist you. We look forward to seeing you there! Join us ：https://discord.gg/raweFPmdzG
Thank you again for your contribution❤
Best regards！ @luminxu

PR #391, opened 2020-12-30
compress data size

jin-s13:


innerlee:
Actually this change will increases the total size of the repo, because git remembers all the history.

The spirit of this change is good: making the repo small when possible, so that cloning could be faster.

innerlee (approved):

innerlee:
In future, we can suppress them in pull requests, before being merged.

PR #405, opened 2021-01-05
[Feature] Return heatmap

kennymckormick:
1. Add the interface in `tools/test.py` to return heatmap.
2. Add the interface in `tools/test.py` to deal with simple test sets list[dict(image_path=blahblah, bbox=blahblah)].
3. Store a compact version of heatmap to save space.

### DIFF
--- a/mmpose/core/post_processing/compact_heatmap.py
+++ b/mmpose/core/post_processing/compact_heatmap.py
@@ -1,0 +1,75 @@
+import numpy as np
+
+
+def compact_heatmap(heatmap, threshold=3e-3):
+    """Convert Heatmap into a more compact format. The input heatmap is a numpy
+    arrayof the shape H x W. The function will find a tight rectangle that
+    contain all positions with score > threshold, then crop the heatmap with
+    the tight rectangle. The output heatmap will also be converted to float16.
+    To restore the original heatmap, a quadruple (x, y, original_w, original_h)
+    of type int16 will also be returned.
+
+    Args:
+        heatmap (np.ndarray([H, W])): The input heatmap.
+        threshold (float): The heatmap score threshold.
+
+    Returns:
+        tuple: New Heatmap and Crop Info.
+        - new_heatmap (np.ndarray([h, w])): The cropped heatmap.
+        - quadruple (np.ndarray([4])): The crop info.
+    """
+    y, x = np.where(heatmap > threshold)
+    min_x, max_x = min(x), max(x)
+    min_y, max_y = min(y), max(y)
+    new_heatmap = heatmap[min_y:max_y, min_x:max_x]
+    new_heatmap = new_heatmap.astype(np.float16)
+    h, w = heatmap.shape
+    quadruple = np.array([min_x, min_y, w, h]).astype(np.int16)
+    return new_heatmap, quadruple
+
+
+def compact_heatmaps(heatmaps, threshold=3e-3):
+    """Convert Heatmap into a more compact format. The input heatmap is a numpy
+    arrayof the shape C x H x W. The function apply compact_heatmap to each
+    channel and return a list of results.
+
+    Args:
+        heatmaps (np.ndarray([C, H, W])): The input heatmap.
+        threshold (float): The heatmap score threshold.
+
+    Returns:
+        list[tuple]: New Heatmaps and Crop Infos.
+    """
+    results = []
+    for heatmap in heatmaps:
+        results.append(compact_heatmap(heatmap, threshold))
+    return results
+
+
+def recover_compact_heatmap(compact_heatmaps):
+    """Convert Compact Heatmap into normal format. The input compact_heatmap is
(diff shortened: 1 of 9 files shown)


jin-s13:
This is only used for Top-down pose estimation. Right?

tools/test.py
```diff
@@ -73,7 +83,14 @@ def main():
         distributed = True
         init_dist(args.launcher, **cfg.dist_params)
 
+    if args.simple_inference:
+        cfg.data.test.simple_inference = True
+
     # build the dataloader
+    # TODO: support multiple images per gpu (only minor changes are needed)
+    if args.simple_inference:
+        cfg.data.test.simple_inference = True
```
jin-s13, line 92: Duplicated?

jin-s13:
@wusize  Please check if this change will conflict with the multi-batch inference.

mmpose/models/detectors/top_down.py
```diff
@@ -185,9 +185,16 @@ def forward_test(self, img, img_metas, return_heatmap=False, **kwargs):
...
         output = self.backbone(img)
 
-        return self.process_head(
+        all_preds, all_boxes, image_path, heatmap = self.process_head(
             output, img, img_metas, return_heatmap=return_heatmap)
 
+        # convert the dtype of heatmap to np.float16 and only keep value > 3e-3
+        # (to save space)
+        if heatmap is not None:
+            heatmap = compact_heatmaps(heatmap[0])
```
jin-s13, line 194: What if the batchsize > 1?

tools/test.py
kennymckormick: Yes, removed.

mmpose/models/detectors/top_down.py
kennymckormick: We do not support batch size > 1 now because:
At test time, the batch size is 1 (hardcoded in `tools/test.py`). The code will be executed if and only if we use `tools/test.py` and set return_heatmap as True.
I will add an assertion line here to make the condition explicit.

kennymckormick:
> This is only used for Top-down pose estimation. Right?

Yes, only support Top-down now.

mmpose/models/detectors/top_down.py
jin-s13: Not really. In v0.10.0, we have already supported multi-batch inference (batch_size>1).

https://github.com/open-mmlab/mmpose/blob/3fe69444d1cced87bfaffaeb20da328177d10148/configs/top_down/resnet/coco/res50_coco_256x192.py#L122

kennymckormick: OK, I will update the code to support multi-batch testing

mmpose/core/post_processing/compact_heatmap.py
```diff
@@ -0,0 +1,82 @@
...
+    h, w = heatmap.shape
+    y, x = np.where(heatmap > threshold)
+
+    # In which case not a single element is larger than threshold
+    if x.shape[0] == 0:
+        new_heatmap = np.zeros([0, 0], dtype=np.float16)
+        return new_heatmap, np.array([0, 0, w, h], dtype=np.int16)
+    min_x, max_x = min(x), max(x)
+    min_y, max_y = min(y), max(y)
+    new_heatmap = heatmap[min_y:max_y, min_x:max_x]
```
innerlee, line 31: Double check if this is needed: `max_y` -> `max_y+1`

mmpose/core/post_processing/compact_heatmap.py
```diff
@@ -0,0 +1,82 @@
+import numpy as np
+
+
+def compact_heatmap(heatmap, threshold=3e-3):
+    """Convert Heatmap into a more compact format. The input heatmap is a numpy
+    arrayof the shape H x W. The function will find a tight rectangle that
```
innerlee, line 6: array of

mmpose/datasets/datasets/bottom_up/bottom_up_coco.py
```diff
@@ -46,14 +47,20 @@ class BottomUpCocoDataset(BottomUpBaseDataset):
         pipeline (list[dict | callable]): A sequence of data transforms.
         test_mode (bool): Store True when building test or
             validation dataset. Default: False.
+        simple_inference (bool): If running the simple inference mode. In
```
innerlee, line 50: How about implementing this as a new dataset, say `BottomUpSimpleDataset`

mmpose/core/post_processing/compact_heatmap.py
kennymckormick: It's OK to use `max_y` here since the shape can always match (we just record the left-up corner of the compact heatmap and restore the heatmap with the coordinates).

kennymckormick:
This PR is stale, maybe I will reopen a new one later.

innerlee:
Okay

PR #411, opened 2021-01-08
UDP speed up.

yaochaorui:


innerlee:
Any numbers on speed comparison?

jin-s13 (approved):

yaochaorui:
> Any numbers on speed comparison?


Sorry for late reply. Yes. For fair comparison, I simply compare the two, the original and the modified, in a single run with same data. To get more robust estimation, I repeated 10000 times in a single run. 

Data:
0.024538049031980337 ms
0.4028586824890226 ms

I did not attach the result in that by simple mental experiments, it is easy to verify the speed-up.  It is more convincing by data.

PR #423, opened 2021-01-15
set add_keypoint_weight=False by default, to speed up visualization.

jin-s13:


mmpose/models/detectors/bottom_up.py
```diff
@@ -303,6 +303,7 @@ def show_result(self,
                     font_scale=0.5,
                     win_name='',
                     show=False,
+                    add_keypoint_weight=False,
```
innerlee, line 306: `add_keypoint_weight` -> `show_keypoint_weight` or `draw_keypoint_weight`?

innerlee (approved):

PR #427, opened 2021-01-17
PoseTracking demo with MMTracking

jin-s13:
Add a small video (200k) to run demos.
Shall we include MMTracking in the requirements?


innerlee:
Optional dependency is ok

mmpose/models/detectors/mesh.py
```diff
@@ -1,12 +1,17 @@
...
-from smplx import SMPL
 
 from .. import builder
 from ..mesh_heads.discriminator import SMPLDiscriminator
 from ..registry import POSENETS
 from .base import BasePose
 
+try:
+    from smplx import SMPL
+    with_smpl = True
```
innerlee, line 11: with_smpl -> has_smpl

innerlee (approved):

PR #434, opened 2021-01-23
fix a bug in apis/train.py

wusize:
fix a bug in apis/train.py


jin-s13 (approved):

innerlee:
drop last in training is not uncommon, so i suggest making it an configuratable option

innerlee:
My mistake, it is for val

PR #441, opened 2021-01-25
Fix bugs in sigmas of AIC

jin-s13:


mmpose/datasets/datasets/top_down/topdown_aic_dataset.py
```diff
@@ -78,11 +78,13 @@ def __init__(self,
             [1., 1.2, 1.5, 1., 1.2, 1.5, 1., 1.2, 1.5, 1., 1.2, 1.5, 1., 1.],
             dtype=np.float32).reshape((self.ann_info['num_joints'], 1))
 
+        # 'https://github.com/AIChallenger/AI_Challenger_2017/blob/master/'
+        # 'Evaluation/keypoint_eval/keypoint_eval.py#L50'
```
innerlee, line 82: add reason for the removal of `2 *`

jin-s13: # note: delta equals to two times of sigma

innerlee (approved):

PR #469, opened 2021-02-07
add albu

jin-s13:


mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -142,3 +150,164 @@ def __repr__(self):
...
+    def __repr__(self):
+        repr_str = self.__class__.__name__
+        repr_str += f'(mean={list(self.mean)}, '
+        repr_str += f'std={list(self.std)}, '
+        repr_str += f'to_rgb={self.to_rgb})'
+        return repr_str
+
+
+@PIPELINES.register_module()
+class Albu(object):
```
innerlee, line 188: Is Albu the standard abbreviation of Albumentation?

irvingzhang0512:
Just curious, why does this pr support pixel-level transforms only? I remember albu could augment images and keypoints and the same time. Any insights about this?

jin-s13:
@irvingzhang0512 Yes. Albumentation also supports spatial-level transforms, see [this](https://albumentations.ai/docs/getting_started/keypoints_augmentation/).
Currently, MMPose does not support spatial-level transforms for the following reasons.

(1) Some spatial augmentations may affect class labels and make them incorrect. For example, the HorizontalFlip augmentation mirrors the input image. And the joint id should also be flipped (left wrist <--> right wrist).  And Albumentation cannot  handle this case automatically. 
(2) I did not check all the transforms. I am not sure if they will have some unexpected behaviors.
(3) Besides 2d keypoint localization, MMPose also supports other tasks, e.g. 3d keypoint localization tasks. Their targets should also be changed accordingly. However, not all of them are supported by Albumentation. 

To make it **safer**, I consider just supporting pixel-level transforms (which will not affect targets) **for now**.

Do you have any ideas?

innerlee:
In mmaction2, there are similar issues for optical flow. These packages

mmpose/datasets/pipelines/shared_transform.py
jin-s13: Maybe the standard abbreviation of Albumentation is 'A'?  : )
It reads `import albumentations as A` in their official website.

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -142,3 +150,164 @@ def __repr__(self):
         """Compute the string representation."""
         return (f'{self.__class__.__name__}('
                 f'keys={self.keys}, meta_keys={self.meta_keys})')
+
+
+@PIPELINES.register_module()
+class Normalize(object):
```
innerlee, line 156: Is Normalize used anywhere?

jin-s13: Not used for now. 
It looks like a standard pipeline, so I also added it.

innerlee: Let's add it later.

Currently all normalization comes after the ToTensor, using NormalizeTensor.
Either we keep this convention, or we mv all normalization before ToTensor to use Normalize.

jin-s13: OK.

mmpose/datasets/pipelines/shared_transform.py
innerlee: That's too much haha

innerlee: How about "Albumentation“

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -142,3 +149,132 @@ def __repr__(self):
         """Compute the string representation."""
         return (f'{self.__class__.__name__}('
                 f'keys={self.keys}, meta_keys={self.meta_keys})')
+
+
+@PIPELINES.register_module()
+class Albu(object):
+    """Albumentation augmentation (pixel-level transforms only) Adds custom
```
innerlee, line 156: Add period before `Adds`, and remove comma before `visit`

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -142,3 +149,132 @@ def __repr__(self):
...
+                ],
+                p=0.1),
+        ]
+    Args:
+        transforms (list[dict]): A list of albu transformations
+        keymap (dict): Contains {'input key':'albumentation-style key'}
+    """
+
+    def __init__(self, transforms, keymap=None):
+        self.allowed_metrics = [
```
innerlee, line 188: metrics?

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -142,3 +149,132 @@ def __repr__(self):
...
+                'img': 'image',
+            }
+        else:
+            self.keymap_to_albu = keymap
+        self.keymap_back = {v: k for k, v in self.keymap_to_albu.items()}
+
+    def albu_builder(self, cfg):
+        """Import a module from albumentations.
+
+        It inherits some of :func:`build_from_cfg` logic.
```
innerlee, line 221: inherits -> resembles

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -142,3 +149,132 @@ def __repr__(self):
...
+            dict(
+                type='RandomBrightnessContrast',
+                brightness_limit=[0.1, 0.3],
+                contrast_limit=[0.1, 0.3],
+                p=0.2),
+            dict(type='ChannelShuffle', p=0.1),
+            dict(
+                type='OneOf',
+                transforms=[
+                    dict(type='Blur', blur_limit=3, p=1.0),
```
innerlee, line 177: Can the transforms be nested?

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -142,3 +149,132 @@ def __repr__(self):
...
+        Renames keys according to keymap provided.
+        Args:
+            d (dict): old dict
+            keymap (dict): {'old_key':'new_key'}
+        Returns:
+            dict: new dict.
+        """
+
+        updated_dict = {}
+        for k, v in zip(d.keys(), d.values()):
```
innerlee, line 263: `zip(d.keys(), d.values())` -> `d.items()`?

innerlee: This could be a one-liner

```python
updated_dict = {keymap.get(k, k):v for k, v in d.items()}
```

mmpose/datasets/pipelines/shared_transform.py
jin-s13: OK.

mmpose/datasets/pipelines/shared_transform.py
jin-s13: changed to transforms

mmpose/datasets/pipelines/shared_transform.py
jin-s13: What do you mean by nested?

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -142,3 +149,129 @@ def __repr__(self):
...
+                type='OneOf',
+                transforms=[
+                    dict(type='Blur', blur_limit=3, p=1.0),
+                    dict(type='MedianBlur', blur_limit=3, p=1.0)
+                ],
+                p=0.1),
+        ]
+    Args:
+        transforms (list[dict]): A list of albu transformations
+        keymap (dict): Contains {'input key':'albumentation-style key'}
```
innerlee, line 184: Give a concrete example for keymap

mmpose/datasets/pipelines/shared_transform.py
innerlee: don't forget about the renaming

mmpose/datasets/pipelines/shared_transform.py
innerlee: Nevermind, the implementation is recursive

mmpose/datasets/pipelines/shared_transform.py
jin-s13: OK.

innerlee (approved):

PR #481, opened 2021-02-19
add decode

liuxin9608:


mmpose/models/keypoint_heads/fc_head.py
```diff
@@ -115,6 +115,25 @@ def inference_model(self, x, flip_pairs=None):
...
+        Args:
+            img_metas (list(dict)): Information about data augmentation
+                By default this includes:
+                - "image_file: path to the image file
+                - "center": center of the bbox
+                - "scale": scale of the bbox
+                - "rotation": rotation of the bbox
+                - "bbox_score": score of bbox
+            output (np.ndarray[N, K, H, W]): model predicted heatmaps.
+            kwargs: dict contains 'img_size'.
```
innerlee, line 130: Why not just add `img_size` to argument list?

mmpose/models/keypoint_heads/fc_head.py
```diff
@@ -115,6 +115,25 @@ def inference_model(self, x, flip_pairs=None):
             output_regression = output.detach().cpu().numpy()
         return output_regression
 
+    def decode(self, img_metas, output, **kwargs):
```
innerlee, line 118: decode_heatmap?

jin-s13: We will support multi-task network with multiple heads for depth estimation, classification, and keypoint estimation soon. To make a unified interface for different heads, we will add a `decode` function for all types of heads. 

For keypoint head, `decode` means `decode_heatmap`; while for depth head, `decode` means `decode_depthmap` etc.

innerlee: How to determine which `decode_xx` to call?

mmpose/models/keypoint_heads/fc_head.py
innerlee: Okay this is part of interface refactoring

mmpose/models/keypoint_heads/fc_head.py
innerlee: I'm trying to understand the design philosophy here. So if you could provide more details...

innerlee: Okay, as discussed offline, we rename all decode_xx into decode. There will be no multiple decodes in one class.

mmpose/models/keypoint_heads/fc_head.py
innerlee: Still not sure about this. Leave it to future revision

innerlee (approved):

PR #482, opened 2021-02-20
Wingloss

jin-s13:


configs/face/deeppose/wflw/deeppose_res50_wflw_256x256.py
```diff
@@ -46,11 +46,7 @@
         num_joints=channel_cfg['num_output_channels'],
         loss_keypoint=dict(type='SmoothL1Loss', use_target_weight=True)),
     train_cfg=dict(),
-    test_cfg=dict(
-        flip_test=True,
-        post_process='default',
-        shift_heatmap=True,
-        modulate_kernel=11))
+    test_cfg=dict())
```
innerlee, line 49: Reason and effect of this change?

mmpose/models/losses/regression_loss.py
```diff
@@ -20,18 +29,72 @@ def forward(self, output, target, target_weight):
...
+                 loss_weight=1.):
+        super().__init__()
+        self.omega = omega
+        self.epsilon = epsilon
+        self.use_target_weight = use_target_weight
+        self.loss_weight = loss_weight
+
+    def criterion(self, pred, target):
+        delta = (target - pred).abs()
+        c = self.omega * (1.0 - math.log(1.0 + self.omega / self.epsilon))
```
innerlee, line 74: It seems that `c` could be pre-computed

mmpose/models/losses/regression_loss.py
```diff
@@ -20,18 +29,72 @@ def forward(self, output, target, target_weight):
...
+        losses = torch.where(
+            torch.lt(delta, self.omega),
+            self.omega * torch.log(1.0 + delta / self.epsilon), delta - c)
```
innerlee, line 77: `self.omega * torch.log(1.0 + delta / self.epsilon)` could be cached also

mmpose/models/losses/regression_loss.py
```diff
@@ -20,18 +29,72 @@ def forward(self, output, target, target_weight):
...
+        self.omega = omega
+        self.epsilon = epsilon
+        self.use_target_weight = use_target_weight
+        self.loss_weight = loss_weight
+
+    def criterion(self, pred, target):
+        delta = (target - pred).abs()
+        c = self.omega * (1.0 - math.log(1.0 + self.omega / self.epsilon))
+        losses = torch.where(
+            torch.lt(delta, self.omega),
```
innerlee, line 76: use `<` instead of `lt`

mmpose/models/losses/regression_loss.py
```diff
@@ -20,18 +29,72 @@ def forward(self, output, target, target_weight):
...
+            target_weight (torch.Tensor[N, K, D]):
+                Weights across different joint types.
+        """
+        if self.use_target_weight:
+            loss = self.criterion(output * target_weight,
+                                  target * target_weight)
+        else:
+            loss = self.criterion(output, target)
+
+        return loss * self.loss_weight
```
innerlee, line 46: Looks like that the implementation of `SmoothL1Loss` is changed.

Why remove the ` / num_joints` part?

configs/face/deeppose/wflw/deeppose_res50_wflw_256x256.py
jin-s13: They are not used in fc_head.

mmpose/models/losses/regression_loss.py
jin-s13: Yes. 
In nn.SmoothL1Loss, reduction='mean', so the loss has already been divided by the number of elements in the output. No need to divide `num_joints` again.

innerlee: Okay, does this change affect any ckpt?

jin-s13: Yes. This may affect the ckpts of deeppose.

mmpose/models/losses/regression_loss.py
jin-s13: Yes.

tests/test_loss/test_regression_losses.py
```diff
@@ -63,3 +63,6 @@ def test_wing_loss():
     assert torch.gt(
         loss(fake_pred, fake_label, torch.ones_like(fake_label)),
         torch.tensor(.5))
+
+
+test_wing_loss()
```
innerlee, line 68: This line is not needed in the final version.

BTW for better experience in debugging unittest, try vscode. 

image

mmpose/models/losses/regression_loss.py
jin-s13: I retrained the model, but did not notice significant difference in accuracy.

innerlee (approved):

PR #495, opened 2021-02-24
fix num_factors

### LINKED ISSUE #494: Confusion about udp target generation
[Code](https://github.com/open-mmlab/mmpose/blob/13d44d09260d95880b3aa504f93441eaab1ee347/mmpose/datasets/pipelines/top_down_transform.py#L559) from `top_down_transform.py`
```python
        elif self.encoding == 'UDP':
            if self.target_type == 'CombinedTarget':
                factors = self.valid_radius_factor
                channel_factor = 3
            elif self.target_type == 'GaussianHeatMap':
                factors = self.sigma
                channel_factor = 1
            if isinstance(factors, list):
                num_factors = len(self.kernel)
                cfg = results['ann_info']
                num_joints = cfg['num_joints']
                W, H = cfg['heatmap_size']

                target = np.empty((0, channel_factor * num_joints, H, W),
                                  dtype=np.float32)
                target_weight = np.empty((0, num_joints, 1), dtype=np.float32)
                for i in range(num_factors):
                    target_i, target_weight_i = self._udp_generate_target(
                        cfg, joints_3d, joints_3d_visible, factors[i],
                        self.target_type)
                    target = np.concatenate([target, target_i[None] * 255], axis=0)
                    target_weight = np.concatenate(
                        [target_weight, target_weight_i[None]], axis=0)
            else:
                target, target_weight = self._udp_generate_target(
                    results['ann_info'], joints_3d, joints_3d_visible, factors,
                    self.target_type)
```
If `factor` is not a list, it's clear that `factor` is either `sigma` or `valid_radius_factor`.
But if `factor` is a list, the length of factor, `num_factors`, is determined by the length of the kernel.
Kernel should have nothing to do with generating udp target, right?
Is there any specific reason for this?

jin-s13:
fix #494

### DIFF
--- a/mmpose/datasets/pipelines/top_down_transform.py
+++ b/mmpose/datasets/pipelines/top_down_transform.py
@@ -564,7 +564,7 @@ class TopDownGenerateTarget:
                 factors = self.sigma
                 channel_factor = 1
             if isinstance(factors, list):
-                num_factors = len(self.kernel)
+                num_factors = len(factors)
                 cfg = results['ann_info']
                 num_joints = cfg['num_joints']
                 W, H = cfg['heatmap_size']


innerlee:
Old question: any ckpt affected?

jin-s13:
> Old question: any ckpt affected?

No.

innerlee (approved):

PR #504, opened 2021-02-26
change channel order according to cfg

liuxin9608:


demo/top_down_video_demo_with_mmdet.py
```diff
@@ -83,7 +84,10 @@ def main():
 
     dataset = pose_model.cfg.data['test']['type']
 
-    cap = cv2.VideoCapture(args.video_path)
+    if args.video_path != '':
+        cap = cv2.VideoCapture(args.video_path)
+    else:
+        cap = cv2.VideoCapture(0)
```
innerlee, line 90: Is this for live camera?

liuxin9608: yes

innerlee: Refer to https://github.com/open-mmlab/mmaction2/blob/master/demo/webcam_demo.py for an implementation of webcam. It uses two threads for capturing and analyzing, which improves user experience

innerlee: So I suggest writing a separate file for webcam, and make this pr focus on channel order

mmpose/apis/inference.py
```diff
@@ -117,18 +118,23 @@ def __call__(self, results):
...
         if isinstance(results['img_or_path'], str):
             results['image_file'] = results['img_or_path']
-        else:
+            img = mmcv.imread(results['img_or_path'], self.color_type,
+                              self.channel_order)
+        elif isinstance(results['img_or_path'], np.ndarray):
             results['image_file'] = ''
-        img = mmcv.imread(results['img_or_path'], self.color_type,
-                          self.channel_order)
+            if self.channel_order == 'rgb':
```
innerlee, line 132: add check for color_type for this if statement

demo/top_down_video_demo_with_mmdet.py
jin-s13: agree

innerlee (approved):

PR #525, opened 2021-03-20
Add checkpoint and config files

jin-s13:
1. Enrich the modelzoo. (more checkpoints and config files.)
2. Use hasattr to exclude unsupported transforms in albumentations.
3. Fix init_weights() in vgg.py


innerlee (approved):

innerlee:
Note for future: unrelated changes should be put in a separate pr :)

PR #527, opened 2021-03-22
Preprocess h36m

cherryjm:


tools/dataset/preprocess_h36m.py
```diff
@@ -0,0 +1,347 @@
...
+        assert len(centers) == len(imgnames)
+
+        print('Annoatations for sequence \"%s %s\" are loaded.'
+              '%d samples in total.' % (subject, basename, len(imgnames)))
+
+        return imgnames, centers, scales, kps_2d, kps_3d
+
+
+if __name__ == '__main__':
+    METADATA = '/data/Human3.6M/metadata.xml'
```
innerlee, line 336: Please use argparse to deal with customizable paths.

tools/dataset/preprocess_h36m.py
```diff
@@ -0,0 +1,347 @@
...
+    ORIGINAL_DIR = '/data/Human3.6M/original'
+    EXTRACTED_DIR = '/data/Human3.6M/extracted'
```
innerlee, line 338: EXTRACTED_DIR -> EXTRACT_DIR

The folder name does not have to be changed.

tools/dataset/preprocess_h36m.py
```diff
@@ -0,0 +1,347 @@
...
+        self.scale_factor = 1.2
+
+    def extract_tgz(self):
+        """Extract files from self.extrct_files."""
+        os.makedirs(self.extracted_dir, exist_ok=True)
+        for subject in self.subjects_annot:
+            cur_dir = join(self.original_dir, subject.lower())
+            for file in self.extract_files:
+                filename = join(cur_dir, file + '.tgz')
+                print('Extracting %s ...' % filename)
```
innerlee, line 93: f-string

tools/dataset/preprocess_h36m.py
cherryjm: extract and process in the same folder?

innerlee: nonono, ignore it

innerlee: I was uneasy with the variable name `EXTRACTED` because they are to be extracted and are not extracted yet... But nevermind

tools/dataset/preprocess_h36m.py
```diff
@@ -0,0 +1,347 @@
...
+                    tar.extractall(self.extracted_dir)
+        print('Extraction done.\n')
+
+    def generate_cameras_file(self):
+        """Generate cameras.pkl which contains camera parameters for 11
+        subjects each with 4 cameras."""
+        cameras = {}
+        for subject in range(1, 12):
+            for camera in range(4):
+                key = ('S%d' % subject, self.camera_ids[camera])
```
innerlee, line 104: f-str

tools/dataset/preprocess_h36m.py
```diff
@@ -0,0 +1,347 @@
...
+                cameras[key] = self._get_camera_params(camera, subject)
+
+        out_file = self.processed_dir + '/cameras.pkl'
```
innerlee, line 107: use os.path.join

innerlee:
As discussed in person, this script should be double checked to make sure that it is compatible with the previous usage of h3.6m. cc. @zengwang430521

tools/dataset/preprocess_h36m.py
```diff
@@ -0,0 +1,336 @@
...
+                        [0, -np.sin(x), np.cos(x)]])
+        R_y = np.array([[np.cos(y), 0, np.sin(y)], [0, 1, 0],
+                        [-np.sin(y), 0, np.cos(y)]])
+        R_z = np.array([[np.cos(z), np.sin(z), 0], [-np.sin(z),
+                                                    np.cos(z), 0], [0, 0, 1]])
+        R = (R_x @ R_y @ R_z).T
+        T = metadata_slice[3:6].reshape(-1, 1)
+
+        # intrinsics
+        c = metadata_slice[8:10, np.newaxis]
```
innerlee, line 202: np.newaxis -> None since the latter is shorter :)

tools/dataset/preprocess_h36m.py
```diff
@@ -0,0 +1,336 @@
...
+        prefix = join(subject, sub_base, sub_base)
+
+        cap = cv2.VideoCapture(video_path)
+        for i in range(0, num_frames, self.sample_rate):
+            imgname = prefix + f'_{i + 1: 06d}.jpg'
+            imgnames.append(imgname)
+            dest_path = join(self.processed_dir, 'imgs', imgname)
+            if os.path.exists(dest_path):
+                continue
+            cap.set(1, i)
```
innerlee, line 278: Generally speaking, `.set(CAP_PROP_POS_FRAMES, frameNumber)` is not guaratted to be exact. Sequentially decoding all frames and keep the desired ones is more safe.

Btw, does here need exact frame seeking?

tools/dataset/preprocess_h36m.py
```diff
@@ -0,0 +1,336 @@
...
+            success, img = cap.read()
+            if success:
+                cv2.imwrite(dest_path, img)
+        cap.release()
+        imgnames = np.array(imgnames)
+        assert len(centers) == len(imgnames)
+
+        print(
+            f'Annoatations for sequence \"{subject} {basename}\" are loaded. '
```
innerlee, line 287: ` \" ` -> ` " `

tools/dataset/preprocess_h36m.py
```diff
@@ -0,0 +1,336 @@
...
+        imgnames = []
+        video_path = join(subj_dir, 'Videos', basename + '.mp4')
+        sub_base = subject + '_' + basename.replace(' ', '_')
+        img_dir = join(self.processed_dir, 'imgs', subject, sub_base)
+        os.makedirs(img_dir, exist_ok=True)
+        prefix = join(subject, sub_base, sub_base)
+
+        cap = cv2.VideoCapture(video_path)
+        for i in range(0, num_frames, self.sample_rate):
+            imgname = prefix + f'_{i + 1: 06d}.jpg'
```
innerlee, line 273: `f'{prefix}_{i + 1: 06d}.jpg'`

tools/dataset/preprocess_h36m.py
cherryjm: The FPS is 50, so neighboring frames may have little difference. But I think exact frame seeking is better.

innerlee (approved):

PR #529, opened 2021-03-23
Fix unittest

jin-s13:


innerlee:
Need to add docstring for the function, since I cannot guess what's the implication of `is_3d`

innerlee (approved):

PR #535, opened 2021-03-28
speed up post transforms

jin-s13:


mmpose/core/post_processing/post_transforms.py
```diff
@@ -175,22 +175,21 @@ def transform_preds(coords, center, scale, output_size, use_udp=False):
...
         scale_x = scale[0] / (output_size[0] - 1.0)
         scale_y = scale[1] / (output_size[1] - 1.0)
-        target_coords = np.zeros(coords.shape)
-        target_coords[:,
-                      0] = coords[:, 0] * scale_x + center[0] - scale[0] * 0.5
-        target_coords[:,
-                      1] = coords[:, 1] * scale_y + center[1] - scale[1] * 0.5
     else:
-        target_coords = coords.copy()
-        trans = get_affine_transform(center, scale, 0, output_size, inv=True)
```
innerlee, line 191: doc in L149 needs change

mmpose/core/post_processing/post_transforms.py
```diff
@@ -175,22 +175,21 @@ def transform_preds(coords, center, scale, output_size, use_udp=False):
...
-        for p in range(coords.shape[0]):
-            target_coords[p, 0:2] = affine_transform(coords[p, 0:2], trans)
+        scale_x = scale[0] / output_size[0]
+        scale_y = scale[1] / output_size[1]
+
+    target_coords = np.zeros(coords.shape)
```
innerlee, line 189: target_coords will be float64 this way.

Use zeros_like to keep dtype the same as coords

innerlee:
Good to go after two minor comments

innerlee (approved):

Issue #533, opened 2021-03-26
Error while trying to export to ONNX pose model


ccdsad:
**torch 1.6.0
mmpose 0.12.0
mmcv 1.2.7
onnx 1.8.1**
`_python3 tools/pytorch2onnx.py /mmpose/configs/top_down/hrnet/coco/hrnet_w48_coco_256x192.py /mmpose/data/hrnet_w48_coco_256x192-b9e0b3ab_20200708.pth --output-file /mmpose/data/hrnet_w48_coco_256x192-b9e0b3ab_20200708.onnx --show --verify_`

`Use load_from_http loader
2021-03-26 15:30:08,692 - mmpose - WARNING - The model and loaded state dict do not match exactly

[...]

PR #543, opened 2021-03-30
Update pytorch2onnx.py

HoBeom:
#533


innerlee:
CI is failling, check out https://github.com/open-mmlab/mmpose/blob/master/.github/CONTRIBUTING.md#code-style for the linting error

HoBeom:
There seems to be a problem with my account setup. I will commit next time when there is the same issue.

PR #544, opened 2021-03-30
Update mmcv version dependence

ly015:


innerlee:
CI fails

innerlee (approved):

innerlee:
Change torch 1.5.1 to 1.5.0

.github/workflows/build.yml
```diff
@@ -83,23 +84,17 @@ jobs:
...
-          - torch: 1.7.0+cu101
-            torchvision: 0.8.1+cu101
-            mmcv: "cu101/torch1.7.0"
+          - torch: 1.5.1
+            torchvision: 0.6.1
+          - torch: 1.6.0
+            torchvision: 0.7.0
+          - torch: 1.7.0
+            torchvision: 0.8.1
+          - torch: 1.7.0
```
innerlee, line 93: 1.7.0 is repeated

mmpose/__init__.py
```diff
@@ -3,8 +3,8 @@
 
 from .version import __version__, short_version
 
-mmcv_minimum_version = '1.1.3'
-mmcv_maximum_version = '1.3'
+mmcv_minimum_version = '1.2.6'
```
innerlee, line 6: Why is the minimum version changed

ly015: I followed https://github.com/open-mmlab/mmaction2/pull/769/files

ly015: So this should not be changed?

innerlee: Yeah keep the original one until someone reports incompaitibility

.github/workflows/build.yml
innerlee: okay python version is different

PR #548, opened 2021-03-31
fix bugs in bottom-up keypoint score

wusize:
Setting all joint scores to be 0 would cause erroneous results in testing for bottom-up approach


jin-s13 (approved):

innerlee:
Adding a unittest for this error would be great

innerlee:
And, does this change affect any scores in modelzoo?

jin-s13:
This bug is caused by upgrading xtcocotools (in the newer version of xtcocotools, those keypoints with zero-score will be ignored.)

innerlee:
Note for future readers: Please use mmpose >= v0.13.0 if you use xtcocotools >= 1.7

PR #549, opened 2021-04-01
add train_with_stride_conv option to TCN

cherryjm:


mmpose/models/backbones/tcn.py
```diff
@@ -35,6 +37,7 @@ def __init__(self,
                  dropout=0.25,
                  causal=False,
                  residual=True,
+                 optimized_1f=False,
```
ly015, line 40: The argument name and descriptions in the docstring can be more specific. For example, "train_with_stride_conv", would better help the user understand what this argument does.

cherryjm: OK, I'll change that

tests/test_backbones/test_tcn.py
```diff
@@ -57,3 +82,17 @@ def test_tcn_backbone():
...
+        num_blocks=4,
+        kernel_sizes=(3, 3, 3, 3, 3),
+        train_with_stride_conv=True)
+    pose2d = torch.rand((2, 34, 243))
+    feat = model(pose2d)
+    assert len(feat) == 4
+    assert feat[0].shape == (2, 1024, 27)
+    assert feat[1].shape == (2, 1024, 9)
+    assert feat[2].shape == (2, 1024, 3)
+    assert feat[3].shape == (2, 1024, 1)
```
ly015, line 98: Could you please add a test in the unittest to check the model w. or w/o train_with_stride_conv will have the same output and gradient after a forward+backward pass?

cherryjm: Will add a test.

ly015 (approved):

PR #554, opened 2021-04-02
Add pose_lifter detector

ly015:
* add pose lifter detector
* add model regularization register interface
* update TCN and temporal head to support weight norm clip
* update temporal head's decode method to handle global position and normalization
* update body3d_h36m_dataset to evaluate by action categories
* update unittests


mmpose/core/post_processing/post_transforms.py
```diff
@@ -105,6 +105,32 @@ def fliplr_regression(regression,
     return regression_flipped
 
 
+def fliplr_regression_3d(regression, flip_pairs):
```
jin-s13, line 108: Can we just use the above `fliplr_regression`?

mmpose/models/keypoint_heads/temporal_regression_head.py
```diff
@@ -80,14 +110,26 @@ def get_accuracy(self, output, target, target_weight):
             target (torch.Tensor[N, K, 3]): Target keypoints.
             target_weight (torch.Tensor[N, K, 3]):
                 Weights across different joint types.
+            metas (list(dict)): Information about data augmentation.
+                By default this includes:
```
jin-s13, line 114: includes: ... (missing lines)

mmpose/models/keypoint_heads/temporal_regression_head.py
```diff
@@ -41,9 +45,33 @@ def __init__(self,
...
             dict(type='Conv1d'), in_channels, num_joints * 3, 1)
 
+        if self.max_norm is not None:
+            # Apply weight norm clip to conv layers
+            weight_clip = WeightNormClipRegister(self.max_norm)
+            for module in self.modules():
+                if isinstance(module, nn.modules.conv._ConvNd):
+                    weight_clip.register(module)
+
+    def _transform_inputs(self, x):
```
jin-s13, line 55: Why not use the '_transform_inputs()' in `top_down_simple_head.py`?

ly015: The behaviors of these two methods are slightly different. TemporalRegressionHead handles temporal features which may have different dimensions from image features'.

mmpose/models/backbones/tcn.py
```diff
@@ -165,7 +165,7 @@ def __init__(self,
 
         assert num_blocks == len(kernel_sizes) - 1
         for ks in kernel_sizes:
-            assert ks & 2 == 1, 'Only odd filter widths are supported.'
+            assert ks & 1 == 1, 'Only odd filter widths are supported.'
```
innerlee, line 168: & is fancy, but % is more easy to read

tools/dataset/preprocess_h36m.py
```diff
@@ -197,6 +253,8 @@ def _get_camera_params(self, camera, subject):
                                                     np.cos(z), 0], [0, 0, 1]])
         R = (R_x @ R_y @ R_z).T
         T = metadata_slice[3:6].reshape(-1, 1)
+        # convert unit from milimeter to meter
+        T *= 0.001
```
cherryjm, line 257: Will this conflicts with annotations? Since the unit of 3D keypoints has already been converted to meter.

ly015: It won't conflict with annotations. The unit of T needs to be the same as the unit of the 3D keypoint coordinates.

mmpose/core/camera/single_camera.py
```diff
@@ -17,6 +17,7 @@ class SimpleCamera(SingleCameraBase):
         [N, J, C]: shape of a batch of person joint coordinates.
         [N, T, J, C]: shape of a batch of pose sequences.
 
+
```
innerlee, line 20: unintended change?

mmpose/models/keypoint_heads/temporal_regression_head.py
```diff
@@ -20,19 +21,22 @@ class TemporalRegressionHead(nn.Module):
...
+                 max_norm=None,
                  loss_keypoint=None,
                  train_cfg=None,
                  test_cfg=None):
         super().__init__()
 
         self.in_channels = in_channels
         self.num_joints = num_joints
-
+        self.max_norm = max_norm
```
innerlee, line 39: Curious why they do not use spectral norm haha

mmpose/models/keypoint_heads/temporal_regression_head.py
```diff
@@ -41,9 +45,33 @@ def __init__(self,
...
+            inputs (tuple/list of Tensor | Tensor): multi-level img features.
+
+        Returns:
+            Tensor: The transformed inputs
+        """
+        if not isinstance(x, (list, tuple)):
+            return x
+
+        assert len(x) > 0
+        return x[-1]
```
innerlee, line 68: So it simply drops all inputs other than the last one?

mmpose/models/keypoint_heads/temporal_regression_head.py
```diff
@@ -80,14 +110,36 @@ def get_accuracy(self, output, target, target_weight):
...
         target_ = target.detach().cpu().numpy()
-        target_weight_ = target_weight.detach().cpu().numpy()
+        if target_weight is None:
+            target_weight_ = np.ones_like(target_)
+        else:
+            target_weight_ = target_weight.detach().cpu().numpy()
+
+        # Denormalize the predicted pose
+        if 'target_mean' in metas[0] and 'target_std' in metas[0]:
+            target_mean = np.stack([_['target_mean'] for _ in metas])
```
innerlee, line 137: Avoid `_` other than dropping variables. Can use `for m in metas` here

mmpose/models/keypoint_heads/temporal_regression_head.py
ly015: Yes for now. The backbone outputs multi-level features (1D feature pyramid?) but only the top-level is used by the head.

mmpose/core/camera/single_camera.py
ly015: Yeah...will remove it.

configs/body3d/simple_baseline/h36m/linear_h36m_block2x1024_maxnorm.py
```diff
@@ -0,0 +1,136 @@
...
+        num_blocks=2,
+        kernel_sizes=(1, 1, 1),
+        dropout=0.5,
+        max_norm=1.0),
+    keypoint_head=dict(
+        type='TemporalRegressionHead',
+        in_channels=1024,
+        num_joints=16,
+        max_norm=1.0,
+        loss_keypoint=dict(type='MPJPELoss')),
```
cherryjm, line 59: The original TensorFlow implementation uses MSELoss instead of MPJPELoss. Maybe this has some effect on training?

configs/body3d/simple_baseline/h36m/linear_h36m_block2x1024_maxnorm.py
```diff
@@ -0,0 +1,136 @@
+log_level = 'INFO'
+load_from = None
+resume_from = None
+dist_params = dict(backend='nccl')
+workflow = [('train', 1)]
+checkpoint_config = dict(interval=10)
+evaluation = dict(interval=10, metric='mpjpe', key_indicator='MPJPE')
```
innerlee, line 7: How to know when to use the lower case and when to use the upper case, of mpjpe

mmpose/core/utils/regularizations.py
```diff
@@ -0,0 +1,86 @@
+from abc import ABCMeta, abstractmethod, abstractproperty
+
+import torch
+
+
+class PytorchModuleHookRegister(metaclass=ABCMeta):
```
innerlee, line 6: How about the class name `PytorchModuleHook`. I think it is more of a hook, than a register.

mmpose/core/utils/regularizations.py
```diff
@@ -0,0 +1,86 @@
...
+            h = module.register_forward_pre_hook(self.hook)
+        elif self.hook_type == 'backward':
+            h = module.register_backward_hook(self.hook)
+        else:
+            raise ValueError(f'Invalid hook type {self.hook}')
+
+        return h
+
+
+class WeightNormClipRegister(PytorchModuleHookRegister):
```
innerlee, line 55: And WeightNormClipHook

mmpose/datasets/datasets/body3d/body3d_h36m_dataset.py
```diff
@@ -101,15 +94,11 @@ def load_annotations(self):
         data_info = super().load_annotations()
 
         # convert 3D joints from original 24-keypoint to standard 17-keypoint
-        assert data_info['joints_3d'].shape[1] == self.JOINT_NUM_ANNOTATION
-        data_info['joints_3d'] = data_info['joints_3d'][:, self.
-                                                        JOINT_IDX_ANNOTATION]
+        data_info['joints_3d'] = data_info['joints_3d']
```
innerlee, line 97: This change will affect the order of joints, so what the reason?

mmpose/datasets/datasets/body3d/body3d_h36m_dataset.py
```diff
@@ -199,23 +190,31 @@ def evaluate(self, outputs, res_folder, metric='mpjpe', **kwargs):
...
-    def _report_mpjpe(self, keypoint_results):
+    def _report_mpjpe(self, keypoint_results, align_pred_and_gt=True):
         """Keypoint evaluation.
 
         Report mean per joint position error (MPJPE) and mean per joint
         position error after rigid alignment (MPJPE-PA)
+
+        Args:
+            align_pred_and_gt (bool): If true (default), the prediction and the
+                gt will be aligned by subtracting global position (root joint
```
innerlee, line 209: What does `align` mean here, "subtracting global position"?
By alignment, who will be subtracted? pred only, gt only, or both?
A clearer name would be better

mmpose/datasets/datasets/body3d/body3d_h36m_dataset.py
```diff
@@ -224,13 +223,29 @@ def _report_mpjpe(self, keypoint_results):
...
+            action = self._parse_h36m_imgname(
+                self.data_info['imgnames'][target_id])[1]
+            action_category = action.split('_')[0]
+            action_category_indices[action_category].append(idx)
+
         preds = np.stack(preds)
         gts = np.stack(gts)
         masks = np.stack(masks).squeeze(-1) > 0
 
+        if align_pred_and_gt:
```
innerlee, line 235: Okay both subtracts the root. Still, "align" is not a good word

mmpose/models/backbones/tcn.py
```diff
@@ -74,24 +76,27 @@ def __init__(self,
...
+                norm_cfg=norm_cfg))
 
         if residual and in_channels != out_channels:
             self.short_cut = build_conv_layer(conv_cfg, in_channels,
                                               out_channels, 1)
         else:
             self.short_cut = None
 
+        if dropout > 0:
+            self.dropout_layer = nn.Dropout(dropout)
```
innerlee, line 98: self.dropout_layer -> self.dropout

innerlee: Set self.dropout_layer to None if dropout ratio is zero, and use `if self.dropout is None` to determine if there are droput layers

mmpose/models/keypoint_heads/temporal_regression_head.py
innerlee: Okay, add a line of comment like

` # return the top-level feature of the 1D feature pyramid`

We would like to give explanations to every magic number

innerlee:
Mainly some minor comments on naming.

configs/body3d/simple_baseline/h36m/linear_h36m_block2x1024_maxnorm.py
ly015: Key indicators will be changed to lower case in eval_hook.

innerlee: good

mmpose/datasets/datasets/body3d/body3d_h36m_dataset.py
```diff
@@ -224,13 +221,33 @@ def _report_mpjpe(self, keypoint_results):
...
+            action_category_indices[action_category].append(idx)
+
         preds = np.stack(preds)
         gts = np.stack(gts)
         masks = np.stack(masks).squeeze(-1) > 0
 
+        if align_root:
+            # The prediction and groundtruth are aligned by subtracting root
+            # position respectively (equivalent to move the prediction to the
+            # groundtruth root position by translation). Then theroot joint
```
innerlee, line 236: theroot

innerlee (approved):

PR #559, opened 2021-04-07
Animal Pose Estimation

jin-s13:


mmpose/core/evaluation/top_down_eval.py
```diff
@@ -28,6 +28,8 @@ def _calc_distances(preds, targets, mask, normalize):
     """
     N, K, _ = preds.shape
     distances = np.full((N, K), -1, dtype=np.float32)
+    # handle invalid values
+    normalize[np.where(normalize <= 0)] = 1e6
```
innerlee, line 32: use clamp?

innerlee (approved):

innerlee:
One comment above, and

* Change the repo's readme
* Demo demo demo

mmpose/core/evaluation/top_down_eval.py
jin-s13: Note that, I use 1e6 instead of 1e-6 here. 
If `normalize` has invalid values, the normalized distance will be near zero (Any prediction can be regarded as correct.).

innerlee: I see

PR #570, opened 2021-04-13
Refactor 3d head

ly015:
This is to support all evaluation protocols in videopose3d.

### DIFF
--- a/configs/body3d/simple_baseline/h36m/linear_h36m_block2x1024_maxnorm.py
+++ b/configs/body3d/simple_baseline/h36m/linear_h36m_block2x1024_maxnorm.py
@@ -3,9 +3,9 @@ load_from = None
 resume_from = None
 dist_params = dict(backend='nccl')
 workflow = [('train', 1)]
-checkpoint_config = dict(interval=10)
+checkpoint_config = dict(interval=1)
 evaluation = dict(
-    interval=10, metric=['mpjpe', 'p-mpjpe'], key_indicator='mpjpe')
+    interval=1, metric=['mpjpe', 'p-mpjpe'], key_indicator='mpjpe')
 
 # optimizer settings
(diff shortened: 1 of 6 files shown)


cherryjm (approved):

configs/body3d/simple_baseline/h36m/linear_h36m_block2x1024_maxnorm.py
```diff
@@ -77,34 +77,34 @@
     dict(
         type='JointRelativization',
         item='target',
-        visible_item='target_visible',
+        # visible_item='target_visible',
```
innerlee, line 80: Reason of this change

ly015: This was to prevent dimension mismatch between the pose and the visibility after global position restoration in temporal_regression_head.This is fixed in the next commit.

mmpose/datasets/datasets/body3d/body3d_h36m_dataset.py
```diff
@@ -236,12 +236,6 @@ def _report_mpjpe(self, keypoint_results, mode='mpjpe'):
         gts = np.stack(gts)
         masks = np.stack(masks).squeeze(-1) > 0
 
-        # Zero-center the pose around its root joint and remove root
```
innerlee, line 239: To confirm: the relativisation will be done in pipeline, right?

innerlee (approved):

mmpose/datasets/datasets/body3d/body3d_h36m_dataset.py
innerlee: If this affect any configs, please remeber to update them

PR #584, opened 2021-04-19
Swith to fp16 hooks in mmcv

ly015:
* import Fp16OptimizerHook from mmcv if its version>=1.3.1
* add deprecation warning to Fp16OptimizerHook, auto_fp16 and force_fp32

### DIFF
--- a/mmpose/apis/train.py
+++ b/mmpose/apis/train.py
@@ -1,12 +1,24 @@
+import warnings
+
+import mmcv
 import torch
 from mmcv.parallel import MMDataParallel, MMDistributedDataParallel
 from mmcv.runner import DistSamplerSeedHook, EpochBasedRunner, OptimizerHook
+from packaging import version
 
-from mmpose.core import (DistEvalHook, EvalHook, Fp16OptimizerHook,
-                         build_optimizers)
+from mmpose.core import DistEvalHook, EvalHook, build_optimizers
 from mmpose.core.distributed_wrapper import DistributedDataParallelWrapper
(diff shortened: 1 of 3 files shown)


jin-s13 (approved):

mmpose/apis/train.py
```diff
@@ -1,13 +1,25 @@
...
+from mmpose.core import DistEvalHook, EvalHook, build_optimizers
 from mmpose.core.distributed_wrapper import DistributedDataParallelWrapper
 from mmpose.datasets import build_dataloader, build_dataset
 from mmpose.utils import get_root_logger
 
+if version.parse(mmcv.__version__) >= version.parse('1.3.1'):
+    from mmcv.runner import Fp16OptimizerHook
+else:
+    warnings.warn(
+        'Fp16OptimizerHook in mmpose will be deprecateed in the next'
```
innerlee, line 18: deprecated

mmpose/apis/train.py
```diff
@@ -1,13 +1,25 @@
...
+        'release and replaced by Fp16Optimizer provided by mmcv (>=1.3.1)'
```
innerlee, line 19: L15 uses Fp16OptimizerHook as well

mmpose/apis/train.py
ly015: fixed.

mmpose/apis/train.py
ly015: fixed

PR #589, opened 2021-04-19
Add 3dbaseline model

ly015:


mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -9,16 +9,16 @@
 
 
 @PIPELINES.register_module()
-class JointRelativization:
+class ZeroCenterJoint:
```
innerlee, line 12: is `ZeroCenter` a verb? do you have any reference (a link to dictionary, etc.)?

innerlee: Okay

innerlee (approved):

PR #560, opened 2021-04-07
(Speed up TopDown Inference) modified inference_top_down_model, make run on batches of bounding box

namirinz:
I have modify _inference_single_pose_model to for-loop preprocess (center, scale) all bounding boxes (let say N bboxes) and stack them into batch_data. And then feed these batch_data on model to get (N) keypoint at once.

And also modify inference_top_down_model to return output as close as the old one.

PR #592, opened 2021-04-20
Reorganize demos

ly015:
* Reorganize demo folder
* Fix out-of-date mmdet/mmtrack configs
* Fix a bug in inference_top_sown_pose_model which causes track_id missing (seems introduced by #560)

Remaining issues:
* Some video files used in example commands in demos do not exist


innerlee:
I suggest putting scripts directly under demo/ folder. Other resources, including documentations can be organized into subfolders

ly015:
Some videos for demos are not in the repo (e.g. demo_fly_video.avi, demo_macaque.mp4, demo_video.mp4). Are these just arbitrary file names to show the command usage, or we should provide actual video files?

innerlee:
Yes they are not included. @jin-s13 we can provide their download urls maybe

innerlee:
cv2 video capture supports url if I remembered correctly, we can check if the video filenames could be replaced by their links

demo/README.md
```diff
@@ -7,43 +7,43 @@ This page provides tutorials about running demos. Please click the caption for m
 <div align="center">
     <img src="demo_coco.gif" width="600px" alt><br>
```
innerlee, line 8: update img link

innerlee: And img links in our repo's README

innerlee:
the demo paths in https://github.com/open-mmlab/mmpose/blob/master/docs/merge_docs.sh#L11-L21 need to be updated as well

innerlee (approved):

PR #602, opened 2021-04-26
Support Videopose3d

cherryjm:
- add option `pad` to `Body3DBaseDataset`
- add new pipelines & refactor `RelativeJointRandomFlip`
- add `Body3DSemiDataset` for semi-supervised learning
- add trajectory model to pose_lifter for semi-supervised learning
- add semi-supervision loss (warm-up)
- add 'w' and 'h' to H36M cameras.pkl


configs/body3d/videopose/h36m/videopose_h36m_27frames_fullconv_semi-supervised.py
```diff
@@ -71,8 +71,8 @@
         loss_keypoint=dict(type='MPJPELoss', use_target_weight=True)),
     loss_semi=dict(
         type='SemiSupervisionLoss',
-        joint_parents=[0, 0, 1, 2, 0, 4, 5, 0, 7, 8, 9, 8, 11, 12, 8, 14, 15],
-        warmup_epochs=5),
+        joint_parents=[0, 0, 1, 2, 0, 4, 5, 0, 7, 8, 9, 8, 11, 12, 8, 14, 15]),
+    warmup_epochs=5 // 5,
```
jin-s13, line 75: `5 // 5` --> 1

innerlee:
Is this intended?

image

cherryjm:
> Is this intended?
> 
> image

I add image width and height to cameras.pkl, which is used to test pipeline `ImageCoordinateNormalization.`

mmpose/datasets/datasets/body3d/body3d_semi_dataset.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
+from torch.utils.data import Dataset
+
+from mmpose.datasets.builder import DATASETS, build_dataset
+
+
+@DATASETS.register_module()
+class Body3DSemiDataset(Dataset):
```
innerlee, line 8: The `Semi` in `SemiDataset` feels like partial/pseudo/quasi/incomplete. Does semi-supervised have any official abbreviations? Or we use the full version `Body3DSemiSupervisedDataset`

cherryjm: It seems that "semi-supervised" doesn't have an official abbreviation. The full version is better.

mmpose/datasets/datasets/body3d/body3d_base_dataset.py
```diff
@@ -60,6 +60,9 @@ def load_config(self, data_cfg):
         self.seq_len = data_cfg.get('seq_len', 1)
         self.seq_frame_interval = data_cfg.get('seq_frame_interval', 1)
         self.causal = data_cfg.get('causal', True)
+        self.pad = data_cfg.get('pad', False)
+        self.subset = data_cfg.get('subset', 1)
+        self.need_target_2d = data_cfg.get('need_target_2d', False)
```
innerlee, line 65: What does `need_target_2d` mean?

cherryjm: It is used for unlabeled data. we first project the predicted 3D pose  back to 2D image plane, and then calculate the difference from target_2d. Since the input is a sequence of 2D poses, we need to specify the target frame.

innerlee:
This error has been fixed in master. Try to rebase onto master.

image

mmpose/datasets/datasets/body3d/body3d_base_dataset.py
ly015: Consider using "need_2d_label" or "need_2d_groundtruth"  which may be more clear as a config option?

configs/body3d/videopose/h36m/videopose_h36m_1frame_fullconv_supervised_cpn_ft.py
```diff
@@ -0,0 +1,157 @@
...
+# data settings
+data_root = 'data/h36m'
+train_data_cfg = dict(
+    num_joints=17,
+    seq_len=1,
+    seq_frame_interval=1,
+    causal=False,
+    pad=False,
+    joint_2d_src='detection',
+    joint_2d_det_file='/mnt/lustre/yingjiaming/external_files' +
```
ly015, line 71: The data files should be placed under mmpose/data/xxx. Please use a symlink here to the actual file path

mmpose/datasets/datasets/body3d/body3d_base_dataset.py
```diff
@@ -60,6 +60,9 @@ def load_config(self, data_cfg):
         self.seq_len = data_cfg.get('seq_len', 1)
         self.seq_frame_interval = data_cfg.get('seq_frame_interval', 1)
         self.causal = data_cfg.get('causal', True)
+        self.pad = data_cfg.get('pad', False)
```
ly015, line 63: Is this temporal padding? Maybe we can use a more specific term like "temporal_padding".

ly015: We may need to describe the options in data_cfg in the docstring.

mmpose/datasets/datasets/body3d/body3d_h36m_dataset.py
```diff
@@ -80,6 +80,8 @@ def load_config(self, data_cfg):
             self.camera_param = self._load_camera_param(
                 data_cfg['camera_param_file'])
 
+        self.subjects = data_cfg.get('subjects', None)
```
ly015, line 83: The subject/action filter was also added in another PR (https://github.com/open-mmlab/mmpose/blob/feb6d668f351016f6f6a48ce34368fac29e110ce/mmpose/datasets/datasets/body3d/body3d_h36m_dataset.py#L97).

It's okay to leave it and we will solve the conflict later.

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -115,6 +115,141 @@ def __call__(self, results):
...
+            class definition for more details. If None is given, the camera
+            parameter will be obtained during processing of each data sample
+            with the key "camera_param".
+    Required keys:
+        item
+    Modified keys:
+        item
+    """
+
+    def __init__(self, item, camera_param=None):
```
ly015, line 135: It looks a little strange that the class name is "ImageCoordinateNormalization" but the camera_param is needed.
I suggest this transform get h and w from the results. And in the dataset's __getitem__, we put image size information into the fetched sample, from either camera parameters or loaded images (we will extend to RGB based 3D pose estimation in the future)

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -115,6 +115,141 @@ def __call__(self, results):
...
+                          dtype=np.float32)
+        scale = np.array(0.5 * camera_param['w'], dtype=np.float32)
+
+        results[self.item] = (results[self.item] - center) / scale
+
+        return results
+
+
+@PIPELINES.register_module()
+class CameraNormalization:
```
ly015, line 162: If this transform is always bonding with ImageCoordinateNormalization, they should be merged into one transform.

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -115,6 +115,141 @@ def __call__(self, results):
...
+            assert 'camera_param' in results, 'Camera parameters are missing.'
+            camera_param = results['camera_param']
+        assert 'h' in camera_param and 'w' in camera_param
+        assert 'f' in camera_param and 'c' in camera_param
+
+        center = np.array([0.5 * camera_param['w'], 0.5 * camera_param['h']],
+                          dtype=np.float32)
+        scale = np.array(0.5 * camera_param['w'], dtype=np.float32)
+
+        camera_param['f'] = camera_param['f'] / scale
```
ly015, line 198: If "camera_param" is from the class attribute (L190), it's dangerous to modify it in place. Maybe add a deep copy in L190

mmpose/datasets/datasets/body3d/body3d_base_dataset.py
cherryjm: OK. I think "need_2d_label" is better.

mmpose/datasets/datasets/body3d/body3d_base_dataset.py
```diff
@@ -21,6 +21,18 @@ class Body3DBaseDataset(Dataset, metaclass=ABCMeta):
         img_prefix (str): Path to a directory where images are held.
             Default: None.
         data_cfg (dict): config
+            - num_joints: Number of joints.
+            - seq_len: Number of frames in a sequence. Default: 1.
+            - seq_frame_interval: Extract frames from the video at certain
+                intervals. Default: 1.
+            - causal: Whether to use causal convolutions. If set to True, the
```
ly015, line 28: It's better to avoid using the term "causal convolutions" here since it's not about the dataset but the model. I think it's okay to remove the first sentence.

mmpose/datasets/datasets/body3d/body3d_semi_supervision_dataset.py
```diff
@@ -0,0 +1,40 @@
...
+
+from mmpose.datasets.builder import DATASETS, build_dataset
+
+
+@DATASETS.register_module()
+class Body3DSemiSupervisionDataset(Dataset):
+    """Mix Dataset for semi-supervised training in 3D human pose estimation
+    task.
+
+    The dataset combines data from two datasets and
```
ly015, line 12: We can give a more specific description, e.g. the two datasets include a labeled one and an unlabeled one.

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -115,6 +117,110 @@ def __call__(self, results):
...
+        for item in self.item:
+            results[item] = (results[item] - center) / scale
+
+        if self.norm_camera:
+            if self.static_camera:
+                camera_param = self.camera_param
+            else:
+                assert 'camera_param' in results, \
+                    'Camera parameters are missing.'
+                camera_param = copy.deepcopy(results['camera_param'])
```
ly015, line 167: L167 doesn't seem to need a deepcopy, since results['camera_param'] will be overwritten anyway. L163 would need a deepcopy to prevent self.camera_param from being modified.

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -115,6 +117,110 @@ def __call__(self, results):
...
+        if camera_param is None:
+            self.static_camera = False
+        else:
+            self.static_camera = True
+            self.camera_param = camera_param
+        self.need_distortion = need_distortion
+
+    def __call__(self, results):
+        if self.static_camera:
+            camera_param = self.camera_param
```
ly015, line 206: Add deepcopy.

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -235,27 +366,42 @@ def __call__(self, results):
...
                 for left, right in flip_pairs:
                     visible_flipped[..., left, :] = visible[..., right, :]
                     visible_flipped[..., right, :] = visible[..., left, :]
-                results[self.vis_item] = visible_flipped
+                results[vis_item] = visible_flipped
+
+            # flip horizontal distortion coefficients
+            if self.flip_camera:
+                if self.static_camera:
+                    camera_param = self.camera_param
```
ly015, line 393: deepcopy

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -115,6 +117,110 @@ def __call__(self, results):
...
+            self.static_camera = False
+        else:
+            self.static_camera = True
+            self.camera_param = camera_param
+
+    def __call__(self, results):
+        center = np.array(
+            [0.5 * results['image_width'], 0.5 * results['image_height']],
+            dtype=np.float32)
+        scale = np.array(0.5 * results['image_width'], dtype=np.float32)
```
ly015, line 156: Should center/scale be calculated from the image shape or from the human bbox (like the provided ones in h36m data).

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -206,23 +312,48 @@ class RelativeJointRandomFlip:
...
         item
     Modified keys:
-        item
+        item (, camera_param)
     """
 
-    def __init__(self, item, root_index, visible_item=None, flip_prob=0.5):
+    def __init__(self,
+                 item,
+                 flip_cfg,
```
ly015, line 334: Please add a description of "flip_cfg" into the docstring.

mmpose/datasets/pipelines/pose3d_transform.py
cherryjm: The original implementation uses center/scale calculated from the image shape. Human bbox is not used here.

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -235,27 +366,42 @@ def __call__(self, results):
...
+
+            # flip horizontal distortion coefficients
+            if self.flip_camera:
+                if self.static_camera:
+                    camera_param = self.camera_param
+                else:
+                    assert 'camera_param' in results, \
+                        'Camera parameters are missing.'
+                    camera_param = copy.deepcopy(results['camera_param'])
+                assert 'c' in camera_param and 'p' in camera_param
```
ly015, line 398: 'p' should be an optional parameter so we just flip it when it is contained in camera_param.

mmpose/models/detectors/pose_lifter.py
```diff
@@ -20,6 +20,11 @@ def __init__(self,
                  backbone,
                  neck=None,
                  keypoint_head=None,
+                 traj_backbone=None,
```
ly015, line 23: Please add argument descriptions into PoseLifter's docstring.

mmpose/models/detectors/pose_lifter.py
```diff
@@ -29,6 +34,10 @@ def __init__(self,
         self.train_cfg = train_cfg
         self.test_cfg = test_cfg
 
+        # semi-supervised learning
+        self.semi = keypoint_head is not None and traj_head is not None and \
```
ly015, line 38: I suggest using a single condition here, e.g. "self.semi==loss_semi is not None", and check other related arguments with assertions.
BTW, is traj_backbone necessary for semi-supervised learning? What would happen if self.semi==True and traj_backbone is None?

mmpose/models/detectors/pose_lifter.py
```diff
@@ -105,21 +148,56 @@ def forward_train(self, input, target, target_weight, metas, **kwargs):
...
+        # supervised learning
+        # pose model
         features = self.backbone(input)
         if self.with_neck:
             features = self.neck(features)
         if self.with_keypoint:
             output = self.keypoint_head(features)
+        # trajectory model
+        if self.with_traj:
+            traj_features = self.traj_backbone(input)
```
ly015, line 160: need to check with_traj_backbone

ly015: And this part can be put together with L173-L176

mmpose/models/detectors/pose_lifter.py
```diff
@@ -20,6 +20,11 @@ def __init__(self,
                  backbone,
                  neck=None,
                  keypoint_head=None,
+                 traj_backbone=None,
+                 traj_neck=None,
+                 traj_head=None,
+                 loss_semi=None,
+                 warmup_epochs=0,
```
ly015, line 27: I think it's better to move warmup_epochs to the arguments of semi_loss.

mmpose/models/detectors/pose_lifter.py
cherryjm: How about sharing backbone with pose model if traj_backbone is None?

mmpose/models/detectors/pose_lifter.py
```diff
@@ -156,6 +243,13 @@ def forward_dummy(self, input):
         if self.with_keypoint:
             output = self.keypoint_head(features)
 
+        if self.with_traj:
```
ly015, line 246: Please modify L240, L242, and L244: features -> output, to ensure that the variable "output" exists.

mmpose/models/detectors/pose_lifter.py
cherryjm: The loss function is called every iteration. I have no idea how to get the number of iterations in an epoch. It can be changed due to different batch sizes?

mmpose/models/keypoint_heads/temporal_regression_head.py
```diff
@@ -94,10 +94,24 @@ def get_loss(self, output, target, target_weight):
         """
         losses = dict()
         assert not isinstance(self.loss, nn.Sequential)
-        if target_weight is None:
-            target_weight = target.new_ones(target.shape)
-        assert target.dim() == 3 and target_weight.dim() == 3
-        losses['reg_loss'] = self.loss(output, target, target_weight)
+
+        # trajectory model
+        if self.num_joints == 1:
```
ly015, line 99: I suggest we use an argument (e.g. is_trajectory) to explicitly determine whether this head is for trajectory prediction.

mmpose/models/detectors/pose_lifter.py
ly015: Yes, I agree. Please clarify the behavior in the docstring.

mmpose/models/losses/regression_loss.py
```diff
@@ -215,3 +215,121 @@ def forward(self, output, target, target_weight):
...
+        else:
+            loss = torch.mean(
+                torch.abs(output_bone.mean(dim=0) - target_bone.mean(dim=0)))
+
+        return loss * self.loss_weight
+
+
+@LOSSES.register_module()
+class SemiSupervisionLoss(nn.Module):
+    """Semi-supervision loss for unlabeled data. It is composed of projection
```
ly015, line 268: Please add a paper reference.

mmpose/models/losses/regression_loss.py
```diff
@@ -215,3 +215,121 @@ def forward(self, output, target, target_weight):
             loss = self.criterion(output, target)
 
         return loss * self.loss_weight
+
+
+@LOSSES.register_module()
+class BoneLoss(nn.Module):
+    """Bone length loss."""
```
ly015, line 222: Please add a description of "joint_parents".

mmpose/models/keypoint_heads/temporal_regression_head.py
```diff
@@ -94,10 +94,24 @@ def get_loss(self, output, target, target_weight):
...
+            if target.dim() == 2:
+                target.unsqueeze_(1)
+
+            if target_weight is None:
+                target_weight = (1 / target[:, :, 2:]).expand(target.shape)
```
ly015, line 104: Please briefly describe the target_weight for trajectory prediction in the docstring.

configs/body3d/videopose/h36m/videopose_h36m_27frames_fullconv_semi-supervised.py
```diff
@@ -68,11 +68,12 @@
...
         num_joints=1,
-        loss_keypoint=dict(type='MPJPELoss', use_target_weight=True)),
+        loss_keypoint=dict(type='MPJPELoss', use_target_weight=True),
+        is_trajectory=True),
     loss_semi=dict(
         type='SemiSupervisionLoss',
-        joint_parents=[0, 0, 1, 2, 0, 4, 5, 0, 7, 8, 9, 8, 11, 12, 8, 14, 15]),
-    warmup_epochs=1,
+        joint_parents=[0, 0, 1, 2, 0, 4, 5, 0, 7, 8, 9, 8, 11, 12, 8, 14, 15],
+        warmup_iterations=12805),  # num_GPUs = 8
```
jin-s13, line 76: **12805** a magic number?

cherryjm: This is a workaround. Because currently, we cannot access epoch number in loss function. So we use iteration number instead.

innerlee: Can this be implemented by hooks?

ly015: @innerlee @jin-s13 Yes, this can be done by hooks. We plan to add a LossGateHook or something like that, to control the loss backpropagation by conditions like epoch/iteration number. It would be implemented in a future PR and need some design. This is a workaround for this algorithm.

@cherryjm Could you please change it to an equation like warmup_iterations=dataset_size/batch_size/GPU_num, and explain each number in the comment? That would be more clear.

cherryjm: OK. Will modify it.

mmpose/datasets/pipelines/pose3d_transform.py
ly015: I think this could be confusing to users if there are "center" and "scale" in the results from the dataset but we calculate them here again with a different method.

I suggest we add an argument to the h36m_dataset, e.g. "use_person_bbox", to determine the center/scale are from the bounding box (as provided in the annotation) or from the image (as calculated here). And in this transform, it simple get the scale/center from the result. What do you think?

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -206,23 +312,50 @@ class RelativeJointRandomFlip:
...
+        if camera_param is None:
+            self.static_camera = False
+        else:
+            self.static_camera = True
+            self.camera_param = camera_param
+
+        if isinstance(self.item, str):
+            self.item = [self.item]
+        if isinstance(self.flip_cfg, dict):
+            self.flip_cfg = [self.flip_cfg]
```
ly015, line 355: Maybe duplicate the flip_cfg for each item like: self.flip_cfg = [self.flip_cfg] * len(self.item)

ly015, line 355: Maybe duplicate the flip_cfg for each item like: self.flip_cfg = [self.flip_cfg] * len(self.item)

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -206,23 +312,50 @@ class RelativeJointRandomFlip:
     """Data augmentation with random horizontal joint flip around a root joint.
 
     Args:
-        item (str): The name of the pose to flip.
-        root_index (int): Root joint index in the pose.
-        visible_item (str): The name of the visibility item which will be
-            flipped accordingly along with the pose.
+        item (str|list[str]): The name of the pose to flip.
+        flip_cfg (dict|list[dict]): Configurations of the fliplr_regression
```
ly015, line 316: I am afraid this is not clear enough. Here we should either clarify the meaning and data type for each item in flip_cfg, or refer to the docstring of the function "fliplr_regression".

mmpose/datasets/pipelines/pose3d_transform.py
```diff
@@ -235,27 +373,44 @@ def __call__(self, results):
...
+                                                   **self.flip_cfg[i])
 
-            results[self.item] = joints_flipped
+                results[item] = joints_flipped
 
             # flip joint visibility
-            if self.vis_item is not None:
-                assert self.vis_item in results
-                visible = results[self.vis_item]
+            for i, vis_item in enumerate(self.vis_item):
```
jin-s13, line 388: unused `i`?

ly015 (approved):

PR #603, opened 2021-04-26
Animal-pose by Cao et al.

jin-s13:


mmpose/datasets/datasets/__init__.py
```diff
@@ -33,5 +33,5 @@
     'Face300WDataset', 'FaceAFLWDataset', 'FaceWFLWDataset', 'FaceCOFWDataset',
     'Body3DH36MDataset', 'AnimalHorse10Dataset', 'AnimalMacaqueDataset',
     'AnimalFlyDataset', 'AnimalLocustDataset', 'AnimalZebraDataset',
-    'AnimalATRWDataset'
+    'AnimalATRWDataset', AnimalPoseDataset
```
innerlee, line 36: What's the relationship between AnimalPoseDataset and AnimalZebraDataset?

innerlee: We have many dataset classes now, maybe its time to trim them down, and moving from code to configs.
Namely,  if user want to support a new dataset, they do not have to cp/paste/modify a dataset python file, instead they tweak a config file and voilà, works done.

innerlee (approved):

mmpose/datasets/datasets/animal/animal_pose_dataset.py
```diff
@@ -0,0 +1,370 @@
...
+        self.ann_info['use_different_joint_weights'] = False
+        self.ann_info['joint_weights'] = np.array(
+            [
+                1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.2, 1.2, 1.2,
+                1.2, 1.5, 1.5, 1.5, 1.5
+            ],
+            dtype=np.float32).reshape((self.ann_info['num_joints'], 1))
+
+        # Note: The original paper did not provide enough information about
+        # the sigmas. We modified from 'https://github.com/cocodataset/'
```
innerlee, line 98: Can sigmas be computed from data?

jin-s13: No. The sigmas measure the accuracy of the annotation, which should be provided by the dataset.

innerlee: Okay

PR #609, opened 2021-04-28
Bottomup nms

luminxu:
add nms for bottom up dataset and demo.

### DIFF
--- a/demo/bottom_up_img_demo.py
+++ b/demo/bottom_up_img_demo.py
@@ -34,11 +34,6 @@ def main():
     parser.add_argument(
         '--kpt-thr', type=float, default=0.3, help='Keypoint score threshold')
     parser.add_argument(
-        '--use-nms',
-        action='store_true',
-        default=True,
-        help='whether to use pose nms')
-    parser.add_argument(
         '--oks-thr', type=float, default=0.9, help='NMS oks threshold')
 
     args = parser.parse_args()
@@ -71,7 +66,6 @@ def main():
         pose_results, returned_outputs = inference_bottom_up_pose_model(
             pose_model,
             image_name,
-            use_nms=args.use_nms,
             oks_thr=args.oks_thr,
             return_heatmap=return_heatmap,
             outputs=output_layer_names)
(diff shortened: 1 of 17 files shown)


jin-s13 (approved):

mmpose/apis/inference.py
```diff
@@ -492,11 +495,19 @@ def inference_bottom_up_pose_model(model,
...
+            area = (np.max(pred[:, 0]) - np.min(pred[:, 0])) * (
+                np.max(pred[:, 1]) - np.min(pred[:, 1]))
             pose_results.append({
                 'keypoints': pred[:, :3],
+                'score': result['scores'][idx],
+                'area': area,
             })
 
+        # pose nms
+        keep = oks_nms(list(pose_results), oks_thr, sigmas=None)
```
ly015, line 508: pose_results should be list already?

jin-s13: It seems that `list` is not necessary here.

luminxu: I have deleted `list` in inference and (bottom up & top down) dataset files.

ly015 (commented):
Ok the image-level NMS can not be applied in a top-down detector for now. So it's okay to do it in the dataset.

ly015 (approved):

demo/bottom_up_img_demo.py
```diff
@@ -34,7 +34,7 @@ def main():
     parser.add_argument(
         '--kpt-thr', type=float, default=0.3, help='Keypoint score threshold')
     parser.add_argument(
-        '--oks-thr', type=float, default=0.9, help='NMS oks threshold')
+        '--pose-nms-thr', type=float, default=0.9, help='Pose nms threshold')
```
ly015, line 37: I suggest we make it clear that pose-nms-thr is applied on OKS score in the help message.

demo/bottom_up_pose_tracking_demo.py
```diff
@@ -28,26 +28,14 @@ def main():
     parser.add_argument(
         '--kpt-thr', type=float, default=0.5, help='Keypoint score threshold')
     parser.add_argument(
-        '--nms-oks-thr', type=float, default=0.9, help='NMS oks threshold')
+        '--pose-nms-thr', type=float, default=0.9, help='Pose nms threshold')
```
ly015, line 31: I suggest we make it clear that pose-nms-thr is applied on OKS score in the help message.

PR #650, opened 2021-05-18
add trans_prob

jin-s13:


mmpose/datasets/pipelines/top_down_transform.py
```diff
@@ -665,16 +665,20 @@ class TopDownRandomTranslation:
...
 
     def __call__(self, results):
         """Perform data augmentation with random translation."""
         center = results['center']
         scale = results['scale']
-        # reference bbox size is [200, 200] pixels
-        center += self.trans_factor * (2 * np.random.rand(2) - 1) * scale * 200
+        if np.random.rand() <= self.trans_prob:
+            # reference bbox size is [200, 200] pixels
+            center += self.trans_factor * (2 * np.random.rand(2) -
```
innerlee, line 681: `np.random.uniform(-1, 1, size=2)`

innerlee (approved):

ly015:
Do we need to make reference bbox size an argument with a default value (200, 200)?

jin-s13:
> Do we need to make reference bbox size an argument with a default value (200, 200)?

Not necessary for now, I think. And this involves a lot of code changes...

innerlee:
Yeah, currently our tactic is to add comments on every appearance of 200

PR #669, opened 2021-05-25
Switch to MMCV MODEL_REGISTRY

ly015:
* swith to MMCV MODEL_REGISTRY
* change mmcv_minimum_version to 1.3.2
* update CI mmcv version

### DIFF
--- a/mmpose/models/builder.py
+++ b/mmpose/models/builder.py
@@ -1,7 +1,7 @@
-from mmcv.cnn import MODELS as MMCV_MODELS
+from mmcv.cnn import build_model_from_cfg
 from mmcv.utils import Registry
 
-MODELS = Registry('models', parent=MMCV_MODELS)
+MODELS = Registry('models', build_func=build_model_from_cfg)
 
 BACKBONES = MODELS
 NECKS = MODELS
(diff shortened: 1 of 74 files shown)


jin-s13:
`mmpose/models/registry.py` is used in `mmpose/models/detectors/`, deleting it may cause some errors.

ly015:
> `mmpose/models/registry.py` is used in `mmpose/models/detectors/`, deleting it may cause some errors.

Registries are moved into builder.py and the import statements have been modified accordingly.

jin-s13 (approved):

mmpose/models/builder.py
```diff
@@ -1,52 +1,35 @@
-from mmcv.utils import build_from_cfg
-from torch import nn
+from mmcv.cnn import MODELS as MMCV_MODELS
+from mmcv.utils import Registry
 
-from .registry import BACKBONES, HEADS, LOSSES, NECKS, POSENETS
+MODELS = Registry('models', parent=MMCV_MODELS)
```
innerlee, line 4: What's the effect of this `parent`? What happens if we remove this parent?

PR #676, opened 2021-05-27
Fix dependency

RangiLyu:
When installing mmpose with mim, the packages in optional.txt will not be installed. But when I try to import mmpose.apis, the package poseval will be imported. So it should be moved into runtime.txt.


jin-s13:
Thanks. @RangiLyu  Please sign CLA.

innerlee (approved):

ly015 (approved):

PR #681, opened 2021-05-27
Add VideoPose3D README

cherryjm:


jin-s13:
Also add a link in `https://github.com/open-mmlab/mmpose/blob/master/README.md`

innerlee (approved):

PR #683, opened 2021-05-28
Add MPI-INF-3DHP dataset

cherryjm:
- Add preprocessing codes
- Add Body3dMpiInf3dhpDataset class
- add 3DPCK and 3DAUC metrics


tools/dataset/preprocess_mpi_inf_3dhp.py
```diff
@@ -0,0 +1,362 @@
...
+                centers, scales, joints_2d, joints_3d = get_annotations(
+                    joints_2d, joints_3d)
+                _centers.append(centers)
+                _scales.append(scales)
+                _joints_2d.append(joints_2d)
+                _joints_3d.append(joints_3d)
+
+                # extract frames from video
+                video = join(seq_path, 'imageSequence', f'video_{cam}.avi')
+                cap = cv2.VideoCapture(video)
```
ly015, line 175: Could you please try mmcv video IO interface [VideoReader](https://github.com/open-mmlab/mmcv/blob/master/docs/video.md) and mmcv.track_iter_progress (to replace tqdm) here?

mmpose/core/evaluation/pose3d_eval.py
```diff
@@ -50,3 +50,116 @@ def keypoint_mpjpe(pred, gt, mask, alignment='none'):
...
+            - ``'scale'``: align in the least-square sense in scale
+            - ``'procrustes'``: align in the least-square sense in scale,
+                rotation and translation.
+        threshold:  If L2 distance between the prediction and the groundtruth
+            is less then threshold, the predicted result is considered as
+            correct. Default: 0.15 (m).
+    Returns:
+        pck: percentage of correct keypoints.
+    """
+    assert mask.any()
```
ly015, line 86: The "pred" and "gt" are single-person or multi-person keypoints? Is it a possible case that the mask is all False?

cherryjm: They are single-person keypoints. The mask here can't be all False. For multi-person pose estimation, I think maybe we can add a post-processing step to evaluate every single person respectively.

mmpose/datasets/datasets/body3d/body3d_base_dataset.py
```diff
@@ -188,21 +188,20 @@ def prepare_data(self, idx):
...
+            _cam_param = self.get_camera_param(_imgnames[0])
             results['camera_param'] = _cam_param
+            # get image size
+            if 'w' in _cam_param and 'h' in _cam_param:
+                # get image size from camera parameters
+                results['image_width'] = _cam_param['w']
+                results['image_height'] = _cam_param['h']
+            else:
+                # get image size from images
+                raise NotImplementedError
```
ly015, line 204: In this case, the image size will be figured out later in the pipeline, is that? Then maybe we should not raise an exception here.

mmpose/datasets/datasets/body3d/body3d_mpi_inf_3dhp_dataset.py
```diff
@@ -0,0 +1,385 @@
...
+        self.joint_2d_src = data_cfg.get('joint_2d_src', 'gt')
+        if self.joint_2d_src not in self.SUPPORTED_JOINT_2D_SRC:
+            raise ValueError(
+                f'Unsupported joint_2d_src "{self.joint_2d_src}". '
+                f'Supported options are {self.SUPPORTED_JOINT_2D_SRC}')
+
+        self.joint_2d_det_file = data_cfg.get('joint_2d_det_file', None)
+
+        self.need_camera_param = data_cfg.get('need_camera_param', False)
+        if self.need_camera_param:
```
ly015, line 91: Is it allowed for MPI-INF-3DHP to not have camera parameters? If not, consider adding an assertion here.

ly015 (approved):

PR #686, opened 2021-06-01
Use MMCV EvalHook

ly015:
* reimplement EvalHook and DistEvalHook based on mmcv.runner.EvalHook


ly015:
@innerlee I think this PR will not cause BC-breaking. The old-fashioned config is still supported.

ly015:
I submitted an issue to MMCV to suggest using case-insensitive key_indicator (https://github.com/open-mmlab/mmcv/issues/1065). If it is approved, we can simplify our EvalHook implementation.

innerlee:
There is a difference between beaking and compatibility helper. The change replaces an argument with other, so it is breaking. We add some compatibility helper to gently warn the users to update their code. This is another thing.

image

ly015:
This PR has been tested both locally and in a distributed environment.
Please note that the DistEvalHook from mmcv will sync bn buffer by default, which is different from the original DistEvalHook in mmpose. This may lead to different evaluation results, and somehow slightly affect the training process.

jin-s13 (approved):

PR #706, opened 2021-06-07
Support MIM

ly015:
* add MIM introduction in README
* modify setup.py to support mim as mmpose entry point


.gitignore
```diff
@@ -106,6 +106,10 @@ venv.bak/
 
 # custom
 mmpose/version.py
+mmpose/configs
```
jin-s13, line 109: Is it intended?

ly015: Yes. To support MIM, we need to create these symlinks in the folder "mmpose" when installing mmpose, which should not be tracked by git.

ly015: Yes. To support MIM, we need to create these symlinks in the folder "mmpose" when installing mmpose, which should not be tracked by git.

ly015: Alternatively, we can remove these links at the end of setup.py.

jin-s13: It is better to delete them afterwards.

ly015: On second thought, I guess removing the symlinks may cause errors when the mmpose is installed from the source code instead of pip.

jin-s13 (approved):

setup.py
```diff
@@ -99,7 +102,26 @@ def gen_packages_items():
     return packages
 
 
+def link_file_to_package():
+    """Create symlinks to package dependencies to support MIM."""
+    item_list = ['tools', 'configs', 'demo', 'model_zoo.yml']
```
zhouzaida, line 107: model_zoo.yml has been renamed to [model-index.yml](https://github.com/open-mmlab/mmpose/blob/master/model-index.yml), so the 'model_zoo.yml' should be renamed to 'model-index.yml'

setup.py
```diff
@@ -99,7 +104,54 @@ def gen_packages_items():
...
+        # installed by `pip install -e .`
+        mode = 'symlink'
+    elif 'sdist' in sys.argv or 'bdist_wheel' in sys.argv:
+        # installed by `pip install .`
+        # or create source distribution by `python setup.py sdist`
+        mode = 'copy'
+    else:
+        return None
+
+    item_list = ['tools', 'configs', 'demo', 'model-index.yml']
```
zhouzaida, line 126:
```suggestion
    filenames = ['tools', 'configs', 'demo', 'model-index.yml']
```

setup.py
```diff
@@ -99,7 +104,54 @@ def gen_packages_items():
...
+    os.makedirs('mmpose/.mim', exist_ok=True)
```
zhouzaida, line 127:
```suggestion
    mim_root = osp.join('mmpose', '.mim')
    os.makedirs(mim_root, exist_ok=True)
```

setup.py
```diff
@@ -99,7 +104,54 @@ def gen_packages_items():
...
+
+    for item in item_list:
```
zhouzaida, line 129:
```suggestion
    for filename in filenames:
```

setup.py
```diff
@@ -99,7 +104,54 @@ def gen_packages_items():
...
+        if osp.exists(item):
+            tar_path = osp.join('mmpose/.mim', item)
```
zhouzaida, line 131: item renames to filename
```suggestion
            tar_path = osp.join(mim_root, item)
```

setup.py
```diff
@@ -99,7 +104,54 @@ def gen_packages_items():
...
+            elif mode == 'copy':
+                src_path = item
+                if osp.isfile(src_path):
+                    shutil.copyfile(src_path, tar_path)
+                elif osp.isdir(src_path):
+                    shutil.copytree(src_path, tar_path)
+                else:
+                    warnings.warn(f'Cannot copy file {src_path}.')
+
+    return osp.abspath('mmpose/mim')
```
zhouzaida, line 150:
```suggestion
    return osp.abspath('mmpose/.mim')
```

setup.py
```diff
@@ -99,7 +104,54 @@ def gen_packages_items():
...
+
+
 if __name__ == '__main__':
+    mim_path = add_mim_extention()
```
zhouzaida, line 154: Is it necessary to return the mim_path?

innerlee (approved):

setup.py
```diff
@@ -99,7 +104,55 @@ def gen_packages_items():
...
+    """Add extra files that are required to support MIM into the package.
+
+    These files will be added by creating a symlink to the originals if the
+    package is installed in `editable` mode (e.g. pip install -e .), or by
+    copying from the originals otherwise.
+    """
+
+    # parse installment mode
+    if 'develop' in sys.argv:
+        # installed by `pip install -e .`
```
innerlee, line 117: Please check if `pip install -e /path/to/locations/repo`, when cwd is not on the project root, works as intended.

innerlee: I guess L 127 might be affected

ly015: Good suggestion. Fixed in a new commit.

zhouzaida (approved):
LGTM

PR #713, opened 2021-06-10
Fix h36m preproc

ly015:
* Convert focal length "f" unit from mm to m
* Fix docstring in h36m_to_coco.py


innerlee:
Adding a test label for future. No need to add it in this pr

innerlee:
Does this affect any published ckpts?

innerlee (approved):

ly015:
No, checkpoints are safe and sound :)

PR #727, opened 2021-06-16
Body3d video demo

cherryjm:
Add demo for videopose3d


demo/body3d_two_stage_img_demo.py
```diff
@@ -264,7 +264,7 @@ def main():
         vis_3d_pose_result(
             pose_lift_model,
             result=pose_lift_results_vis,
-            img=pose_lift_results[0]['image_name'],
+            img=image_name,
             out_file=out_file)
```
ly015, line 268: What's the purpose of this modification? If it's necessary, please remove the following line.
https://github.com/open-mmlab/mmpose/blob/dc5a4c26578a1edb16c8487971656682c66363fd/demo/body3d_two_stage_img_demo.py#L222

demo/body3d_two_stage_video_demo.py
```diff
@@ -0,0 +1,338 @@
...
+    else:
+        raise NotImplementedError
+
+
+def main():
+    parser = ArgumentParser()
+    parser.add_argument('det_config', help='Config file for detection')
+    parser.add_argument('det_checkpoint', help='Checkpoint file for detection')
+    parser.add_argument(
+        'pose_detector_config',
```
ly015, line 87: Please use the dash "-" for argument names for consistency. And please also help correct the argument names in 3d_two_stage_img_demo.py.

demo/body3d_two_stage_img_demo.py
cherryjm: Because I add a `_gather_pose_lifter_inputs` function to normalize 'keypoints', where only 'keypoints' and 'track_id' are kept . In this way, the only useful information in `pose_lift_results` is 'keypoints_3d'. Other information such as 'image_name' and unnormalized 'keypoints' can be obtained from `pose_det_results`.

jin-s13:
Is the uploaded 'demo/resources/body3d_demo.mp4 ' necessary?

demo/docs/3d_human_pose_demo.md
```diff
@@ -36,3 +36,49 @@ python demo/body3d_two_stage_img_demo.py \
...
+#### Using mmdet for human bounding box detection and top-down model for the 1st stage (2D pose detection), and inference the 2nd stage (2D-to-3D lifting)
+
+Assume that you have already installed [mmdet](https://github.com/open-mmlab/mmdetection).
+
+```shell
+python demo/body3d_two_stage_video_demo.py \
+    ${MMDET_CONFIG_FILE} \
+    ${MMDET_CHECKPOINT_FILE} \
+    ${MMPOSE_CONFIG_FILE_STAGE_1} \
+    ${MMPOSE_CONFIG_FILE STAGE_1} \
```
jin-s13, line 51: CKPT

demo/docs/3d_human_pose_demo.md
```diff
@@ -36,3 +36,49 @@ python demo/body3d_two_stage_img_demo.py \
...
+    ${MMPOSE_CONFIG_FILE_STAGE_2} \
+    ${MMPOSE_CONFIG_FILE STAGE_2} \
```
jin-s13, line 53: CKPT

jin-s13: How about naming it as something like 'MMPOSE_CONFIG_FILE_2D' and 'MMPOSE_CONFIG_FILE_3D'

mmpose/apis/inference_3d.py
```diff
@@ -23,56 +87,81 @@ def _collate_pose_sequence(pose_results, with_track_id=True):
             inner list is the pose information of one person, which contains:
                 keypoints (ndarray[K, 2 or 3]): x, y, [score]
                 track_id (int): unique id of each person, required when
-                    ``with_track_id==True``
-                bbox ((4, ) or (5, )): left, top, right, bottom, [score],
```
jin-s13, line 27: Is `bbox` info removed in pose_results?

mmpose/apis/inference_3d.py
```diff
@@ -6,7 +6,71 @@
...
+
+    Args:
+        pose_results (List[List[Dict]]): Multi-frame pose detection results
+            stored in a nested list. Each element of the outer list is the
+            pose detection results of a single frame, and each element of the
+            inner list is the pose information of one person, which contains:
+                keypoints (ndarray[K, 2 or 3]): x, y, [score]
+                track_id (int): unique id of each person, required when
+                    ``with_track_id==True```
+                bbox ((4, ) or (5, )): left, right, top, bottom, [score]
```
jin-s13, line 29: We still have `bbox` in `pose_results`, right?

cherryjm:
> Is the uploaded 'demo/resources/body3d_demo.mp4 ' necessary?

This is used in the example shell script. I'm not sure whether it is necessary or not. If not, I will delete it.

demo/docs/3d_human_pose_demo.md
cherryjm: Yes, I think that will be more clear.

mmpose/apis/inference_3d.py
cherryjm: This function only keeps keys that are necessary to 2D-3D lifting. So `bbox` is removed from `pose_results` in this function. For visualization, we can get `bbox` from 2D pose detection results.

innerlee:
Do not host it inside repo. You can add a line of wget before running the demo

mmpose/apis/inference_3d.py
jin-s13: OK. I see.

jin-s13 (approved):

demo/body3d_two_stage_video_demo.py
```diff
@@ -0,0 +1,357 @@
...
+    person_results = []
+    for bbox in bboxes:
+        person = {}
+        person['bbox'] = bbox
+        person_results.append(person)
+
+    return person_results
+
+
+def dataset_transform(keypoints, pose_det_dataset, pose_lift_dataset):
```
ly015, line 48: Can we use a more clear function name like "convert_keypoint_format" or "convert_keypoint_definition"?

demo/body3d_two_stage_video_demo.py
```diff
@@ -0,0 +1,357 @@
...
+
+    # transform the predicted 2D keypoints to the keypoints convention of pose
+    # lift dataset
+    for pose_det_results in pose_det_results_list:
+        for res in pose_det_results:
+            keypoints = res['keypoints']
+            res['keypoints'] = dataset_transform(keypoints, pose_det_dataset,
+                                                 pose_lift_dataset)
+
+    # prepare for temporal padding
```
ly015, line 261: Is it possible to wrap temporal padding and index inferring (L261-L276 and L281-288) into helper functions and put them into inference_3d? I think it's too detailed for a demo script.

mmpose/core/visualization/image.py
```diff
@@ -176,7 +176,10 @@ def imshow_keypoints_3d(
         ax_img.imshow(img, aspect='equal')
 
     for idx, res in enumerate(pose_result):
-        kpts = res['keypoints_3d']
+        dummy = len(res) == 0
```
ly015, line 179: Currently, we control the number of visualized instances outside this function, and pass empty pose results as a placeholder, right? 

I suggest we add a function argument like "fix_instance_num" here, and pad or truncate the original pose results during visualization. Do you think it would be more clear?

cherryjm: Yes, I'll modify the visualization function.

ly015:
BTW, it's better to increase the codecov rate to meet the target if possible.

ly015 (approved):

PR #730, opened 2021-06-18
Add back resnet_coco metafile

cherryjm:


innerlee:
What's the meaning of revert?

cherryjm:
The metafile is deleted after reorganizing configs.

innerlee:
Is this done by `git revert`?

jin-s13:
The metafiles should be automatically generated (by codes).

cherryjm:
This is not done by git revert, since the path to configs has changed.
Will add a script later. This PR is just to revert the metafile that has been added before.

ly015:
@cherryjm Could you please also test that mim can access and parse the metafiles properly?

cherryjm:
@ly015 Now the metafiles can be parsed successfully.

ly015 (approved):

PR #745, opened 2021-06-25
rename GaussianHeatMap to GaussianHeatmap

jin-s13:


mmpose/datasets/pipelines/top_down_transform.py
```diff
@@ -253,14 +255,19 @@ def __init__(self,
...
-                 target_type='GaussianHeatMap',
+                 target_type='GaussianHeatmap',
                  encoding='MSRA',
                  unbiased_encoding=False):
         self.sigma = sigma
         self.unbiased_encoding = unbiased_encoding
         self.kernel = kernel
         self.valid_radius_factor = valid_radius_factor
-        self.target_type = target_type
+        if target_type == 'GaussianHeatMap':
```
innerlee, line 265: Ask for opinion: is it safe to compare the lower cased string?

innerlee: If so, we can avoid the deprecation warning

jin-s13: The lower case `gaussianheatmap` looks strange.

innerlee: Yeah, how about we do not use it but we allow it?

innerlee (approved):

mmpose/datasets/pipelines/top_down_transform.py
jin-s13: OK.

ly015:
Please solve the conflict. And it would be great to have a unittest for BC-breaking.

jin-s13:
Both `GaussianHeatMap` and `GaussianHeatmap` work.

PR #761, opened 2021-07-02
add checkpoint and config training on interhand v1.0

zengwang430521:


configs/hand/3d_kpt_sview_rgb_img/internet/interhand3d/internet_interhand3d.md
```diff
@@ -49,7 +49,9 @@ year = {2020}
...
-| All | val(M) | [InterNet_resnet_50](/configs/hand/3d_kpt_sview_rgb_img/internet/interhand3d/res50_interhand3d_all_256x256.py) | 256x256 | 12.03 | 17.88 | 14.84 | 34.93 | 0.99 | [ckpt](https://download.openmmlab.com/mmpose/hand3d/internet/res50_intehand3d_all_256x256-b9c1cf4c_20210506.pth) | [log](https://download.openmmlab.com/mmpose/hand3d/internet/res50_intehand3d_all_256x256_20210506.log.json) |
+|Dataset Version|Train Set| Set | Arch  | Input Size | MPJPE-single |  MPJPE-interacting  |  MPJPE-all  | MRRPE | APh   | ckpt    | log     |
+| :--- | :--- | :--- | :--------: | :--------: | :------: | :------: | :------: |:------: |:------: |:------: |:------: |
+| v0.0 | All | test(H+M) | [InterNet_resnet_50](/configs/hand/3d_kpt_sview_rgb_img/internet/interhand3d/res50_interhand3d_all_256x256.py) | 256x256 | 10.16 | 15.27 | 12.97 | 33.14 | 0.99 | [ckpt](https://download.openmmlab.com/mmpose/hand3d/internet/res50_intehand3d_all_256x256-b9c1cf4c_20210506.pth) | [log](https://download.openmmlab.com/mmpose/hand3d/internet/res50_intehand3d_all_256x256_20210506.log.json) |
```
innerlee, line 54: The four configs are the same, but the checkpoints are different. Is this intended?

zengwang430521: Yes, the network is the same, but trained using different data.

innerlee: If the dataset is different, then the config should be different, I suppose?

zengwang430521: Actually, this is the same dataset. The version v0.0 is first released and I train the first checkpoint using these data. But recently, the version v1.0 is released, so I train another model. 
Maybe, I can remove the first checkpoint and only remain the new one.

innerlee: I see, both can be kept. Add a note in the bottom explains this issue. Also gently remind the user to take care in data preparation.

zengwang430521: I found that the author does not provide the download link of data version 0 now. So, maybe I should remove the previous checkpoint.

innerlee: That's okay then

zengwang430521: fixed.

innerlee (approved):

Issue #509, opened 2021-03-01
Wrong index for PCKh@0.1


AndersonZhangyq:
``` python
rng = np.arange(0, 0.5 + 0.01, 0.01)
array([0.  , 0.01, 0.02, 0.03, 0.04, 0.05, 0.06, 0.07, 0.08, 0.09, 0.1 ,
       0.11, 0.12, 0.13, 0.14, 0.15, 0.16, 0.17, 0.18, 0.19, 0.2 , 0.21,
       0.22, 0.23, 0.24, 0.25, 0.26, 0.27, 0.28, 0.29, 0.3 , 0.31, 0.32,
       0.33, 0.34, 0.35, 0.36, 0.37, 0.38, 0.39, 0.4 , 0.41, 0.42, 0.43,
       0.44, 0.45, 0.46, 0.47, 0.48, 0.49, 0.5 ])
```
The value 0.1 is rng[10], not rng[11]
https://github.com/open-mmlab/mmpose/blob/05cd64c9d4c28e03a35866796e940dc9db1853c8/mmpose/datasets/datasets/top_down/topdown_mpii_dataset.py#L255
should be ```('PCKh@0.1', np.sum(pckAll[10, :] * jnt_ratio))]```

PR #773, opened 2021-07-08
[bug fix] fix mpii pckh 0.1 index

jin-s13:
Also see #516 , #509


innerlee (approved):

innerlee:
Testing this would be great

PR #794, opened 2021-07-19
Add pre-commit hook to update MIM model metafiles and model-index.yml

ly015:
In this PR we:
1. We add a pre-commit hook to call tools/misc/update_model_index.py at a new commit, which will:
    - Update model metafiles according to the corresponding .md files.
    - Update model-index.yml file according to the metafiles.
    - Check if any file has been modified.

2. Organize the `tools` folder


tools/misc/update_model_index.py
```diff
@@ -181,38 +214,47 @@ def parse_md(md_file):
...
+    Returns:
+        Bool: If the updated model-index.yml is different from the original.
+    """
+    configs_dir = osp.join(MMPOSE_ROOT, 'configs')
     yml_files = []
     for root, dirs, files in os.walk(configs_dir):
         if len(dirs) == 0:
             for file in files:
                 if file[-3:] == 'yml':
                     yml_files.append(os.path.join(root, file))
```
innerlee, line 234: Can this discovery logic replaced by glob? like

image

tools/misc/update_model_index.py
```diff
@@ -1,10 +1,47 @@
...
+    Args:
+        obj (any): The python object to be dumped.
+        file (str): YAML filename to dump the object to.
+    Returns:
+        Bool: If the target YAML file is different from the original.
+    """
+
+    original = None
+    if osp.isfile(file):
+        with open(file, 'r') as f:
```
innerlee, line 31: Better to explicitly open with utf8 encoding

tools/misc/update_model_index.py
ly015: Replaced by glob and tested.

tools/misc/update_model_index.py
ly015: fixed.

innerlee (approved):

PR #812, opened 2021-07-26
add pytest.ini and fix docs

jin-s13:


mmpose/models/backbones/litehrnet.py
```diff
@@ -792,15 +789,13 @@ def __init__(self,
                  conv_cfg=None,
                  norm_cfg=dict(type='BN'),
                  norm_eval=False,
-                 with_cp=False,
-                 zero_init_residual=False):
+                 with_cp=False):
```
ly015, line 792: zero_init_residual was unused?

ly015 (approved):

mmpose/models/backbones/litehrnet.py
jin-s13: Not used param.

ly015:
Please test running pytest locally with this pytest.ini

pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
```
innerlee, line 1: no need to add a new file. see https://docs.pytest.org/en/6.2.x/customize.html#setup-cfg

jin-s13: However, here is a warning. 

```
Warning
Usage of setup.cfg is not recommended unless for very simple use cases. .cfg files use a different parser than pytest.ini and tox.ini which might cause hard to track down problems. When possible, it is recommended to use the latter files, or pyproject.toml, to hold your pytest configuration.
```

innerlee: I thought ours belong to the simple case haha. Anyway keep or not, either is okay to me.

PR #814, opened 2021-07-26
[Fix ] Fix docstring and init_weights error of ShuffleNetV1

Junjun2016:
Fix docstring and init_weights error of ShuffleNetV1.


mmpose/models/backbones/shufflenet_v1.py
```diff
@@ -283,7 +283,7 @@ def make_layer(self, out_channels, num_blocks, first_block=False):
             out_channels (int): out_channels of the block.
             num_blocks (int): Number of blocks.
             first_block (bool, optional): Whether is the first ShuffleUnit of a
-                sequential ShuffleUnits. Default: False, which means not using
+                sequential ShuffleUnits. Default: False, which means using
```
jin-s13, line 286: Default: True?

Junjun2016: I just keep the default value of `arg` and correct the docstring to avoid BC-breaking.

jin-s13 (approved):

PR #829, opened 2021-08-02
update mseloss

Ezra-Yu:
1. Add a defaullt value of 'None' for targe_tweight in forward function. If I don't want to use target weight in MSELoss, is is not necessary to pass targte_weight parameters into the forward function.


ly015:
I think this update applies to other losses too. If so, it's better to update them as well in this PR.

jin-s13 (approved):

innerlee:
@ly015 need to add test to check the two arg version

PR #830, opened 2021-08-02
Unify SMPL-like models to mesh models

zengwang430521:


jin-s13 (approved):

mmpose/models/__init__.py
```diff
@@ -1,13 +1,15 @@
 from .backbones import *  # noqa
-from .builder import (BACKBONES, HEADS, LOSSES, NECKS, POSENETS,
-                      build_backbone, build_head, build_loss, build_neck,
-                      build_posenet)
+from .builder import (BACKBONES, EXTERNAL_MODELS, HEADS, LOSSES, NECKS,
```
innerlee, line 2: What does `EXTERNAL_MODELS` mean?

innerlee:
What does "compact" mean?

mmpose/models/__init__.py
jin-s13: Currently, it only contains SMPL. Any suggestions on the name?

innerlee: Something like `MESH_LAYERS`?

ly015: I agree with the name `MESH_LAYERS`.
And I suggest moving the layers as well as the registry and builder into `mmpose/models/utils`.

innerlee (approved):

ly015 (approved):

PR #833, opened 2021-08-03
Remove opencv-python-headless dependency by albumentations

ly015:
By default, installing `albumentations` by pip has `opencv-python-headless` dependency even if `opencv-python` has already been installed. This could cause different opencv versions installed in the environment and lead to unexpected errors.

This PR updates the installation of `albumentations` to prevent using a wheel distribution of it with `opencv-python-headless` dependency. See its [docs](https://albumentations.ai/docs/getting_started/installation/#note-on-opencv-dependencies) for details.

### DIFF
--- a/requirements/optional.txt
+++ b/requirements/optional.txt
@@ -1,4 +1,4 @@
-albumentations>=0.3.2
+albumentations>=0.3.2 --no-binary imgaug,albumentations
 onnx
 onnxruntime
 poseval@git+https://github.com/svenkreiss/poseval.git


jin-s13 (approved):

innerlee:
@ly015 this is an upstream issue. Would you like to ask the albu team to change their dependency to avoid such issues?

ly015:
> @ly015 this is an upstream issue. Would you like to ask the albu team to change their dependency to avoid such issues?

They have a future plan to remove the dependency of opencv-python and use opencv-python-headless only by default, according to [here](https://albumentations.ai/docs/getting_started/installation/#note-on-opencv-dependencies). So I guess the current solution is ok if we hope to keep opencv-python as the single default opencv dependency in mmpose?

innerlee:
That will also cause trouble. Now I don't think they could come up with good solution :(

PR #840, opened 2021-08-06
Update readthedocs settings to support pdf/epub export

ly015:
refer to: https://github.com/open-mmlab/mmdetection/pull/5738/files

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,9 @@
 version: 2
 
+formats: all
+
 python:
-    version: 3.7
-    install:
-      - requirements: requirements/docs.txt
-      - requirements: requirements/readthedocs.txt
+  version: 3.7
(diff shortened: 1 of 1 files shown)


.readthedocs.yml
```diff
@@ -1,7 +1,9 @@
...
+
 python:
-    version: 3.7
-    install:
-      - requirements: requirements/docs.txt
-      - requirements: requirements/readthedocs.txt
+  version: 3.7
+  install:
+    - requirements: requirements/docs.txt
+    - requirements: requirements/readthedocs.txt
```
innerlee, line 9: why not merge the two?

ly015: No idea... Other repos also have these two files. What is the difference between them?

innerlee:
linting fails

PR #849, opened 2021-08-10
add benchmark regression test script with tmux

liqikai9:
### Motivation
When releasing the new version of our codebase monthly or quarterly, we would like to conduct benchmark regression tests for the previously released models and algorithms, which can support different priorities.

The base feature is to read a config file containing a model list and runtime parameters, then run multiple tasks in different panes and windows controlled by [tmux](https://github.com/tmux/tmux) automatically. 

The priority of the models is as follows. P0: core, P1: important, P2: less important, P3: least important. You can assign different priorities for each model and also decide the priority levels for inference and training tasks, respectively.

This script aims at running multiple benchmark regression tasks without the need to start lots of terminals manually and avoiding the possible inconvenience due to network interruption when running tasks on remote servers, which is quite common.

### Modification
We added the folder `.dev_scripts` containing two files:` benchmark_regression_cfg_tmpl.yaml` and `benchmark_regression.py`. Besides, in order to specify the ` work-dir` of the inference task, we added an additional argument  `--work-dir` to the script `$mmpose/tools/test.py`  and modified the code accordingly.

### Arguments
The script is based on `$mmpose/tools/slurm_test.sh`  and `$mmpose/tools/slurm_train.sh`. It supports running  **test** and **train** tasks with custom **priority** and runtime setting parameters, which can be specified in the config file.

To run the script, a config file containing multiple models is required. For example, the `benchmark_regression_cfg_tmpl.yaml` under the directory `$mmpose/.dev_scripts`. The config file gives a template about different fields. It has a `model_list` field that contains different priorities. Under each priority level, there are multiple models. 

Specifically, the config file must indicate model priorities and paths to the config file and the corresponding checkpoint file. For example,
```
model_list:
    P0: # the priority of the models, P0: core, P1: important, P2: less important, P3: least important
      -   config: configs/body/2d_kpt_sview_rgb_img/topdown_heatmap/coco/res50_coco_256x192.py  # path to the config file
          checkpoint: https://download.openmmlab.com/mmpose/top_down/resnet/res50_coco_256x192-ec54d7f3_20200709.pth # path or url to the config file
        task_name: res50_coco_256x192 # the job name in slurm will be specified according to this field and the mode. If not specified, use the basename of the config file

        test:...

        train:...


    -    ...
      
    P1:...
```
The field priority like `P0` and `P1` is added so that you assign different priorities for different models. You can add more models as you need under the corresponding priority field.

For a more detailed description of the arguments, please refer to the script  `$mmpose/.dev_scripts/benchmark_regression.py`.

### Usage
Here is a simple example to run the script.
```
cd $mmpose
python ./.dev_scripts/benchmark_regression.py [--config ${/path/to/moel_list}] [--session-name ${SESSION_NAME}] [--priority ${TEST_PRIORITY} ${TRAIN_PRIORITY} ]
```
Note that the `${TEST_PRIORITY}` and `${TRAIN_PRIORITY}` give the largest number of priorities of test and train tasks, respectively.

Running the above script with default parameters, you will start a new [tmux](https://github.com/tmux/tmux)  session with each pane running a task independently. Enjoy it!


.dev_scripts/config_list.yaml
```diff
@@ -0,0 +1,23 @@
+hrnet_w32_coco_256x192:
```
ly015, line 1: It's better to use a more specific filename like "benchmark_regression_cfg.yaml".

I suggest make the model list a part of the this yaml file, so it can be easily extended to contain new contents in the future.

.dev_scripts/test_benchmark_tmux.py
```diff
@@ -0,0 +1,91 @@
...
+
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    args = parse_args()
+
+    # cd $mmpose
+    os.system('cd {}'.format(args.path))
```
ly015, line 44: We can assume that this script is always called from the mmpose root, so this may not be necessary.

.dev_scripts/test_benchmark_tmux.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # create a new tmux session
+    os.system('tmux new -s {} -d'.format(args.session_name))
```
ly015, line 47: Please always use f-string if possible.

.dev_scripts/test_benchmark_tmux.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # tmux select-window - t 0
+    os.system('tmux select-window -t 0')
+    # pwd
+    os.system('tmux send-keys "%s" "C-m"' % 'pwd')
+    # ls
+    os.system('tmux send-keys "%s" "C-m"' % 'ls')
+
+    with open(args.config) as config_file:
```
ly015, line 56: Consider using `mmcv.load` to avoid introducing the extra dependency on yaml.

.dev_scripts/test_benchmark_tmux.py
```diff
@@ -0,0 +1,91 @@
...
+        '--session_name', '-s', help='the tmux session name', default='test')
+    parser.add_argument(
+        '--path',
+        help='the running path, e.g., $mmpose',
+        default='~/open_mmlab/mmpose')
+    parser.add_argument('--env', help='the conda environment', default='pt1.6')
+    parser.add_argument(
+        '--partition', help='the partition name', default='openmmlab')
+
+    parser.add_argument('--gpus', help='the total number of GPUs', default=2)
```
ly015, line 26: It's better to support loading settings like `gpus`, `gpus-per-node', 'partition', 'cpus_per_task' for each model separately from the config file. And the augments here only give the default values.

.dev_scripts/test_benchmark_tmux.py
```diff
@@ -0,0 +1,91 @@
...
+    parser.add_argument(
+        '--gpus_per_node',
+        default=2,
+        help='the number of GPUs used per computing node',
+        choices=[1, 2, 3, 4, 5, 6, 7, 8])
+    parser.add_argument(
+        '--cpus_per_task', default=2, help='the number of CPUs used per task')
```
ly015, line 33: Please use `-` instead of '_' in the argument name to be consistent with other scripts.

.dev_scripts/test_benchmark_tmux.py
```diff
@@ -0,0 +1,91 @@
...
+    parser.add_argument('--port', default=29666, help='the starting port used')
+
+    args = parser.parse_args()
```
ly015, line 36: Please add an optional argument `--root-work-dir` to specify the path to store logs and checkpoints. If not specified, create a new folder like `work_dirs/benchmark_regression_yyyymmdd' as the root work dir.

.dev_scripts/test_benchmark_tmux.py
```diff
@@ -0,0 +1,91 @@
...
+        # tmux new-window -n hrnet_w32_coco_256x192
+        os.system('tmux new-window -n {}'.format(task_name))
+
+        # tmux select-window - t window_index
+        os.system('tmux select-window -t {}'.format(window_index))
+
+        cmd = 'conda activate {}'.format(args.env)
+        os.system('tmux send-keys "%s" "C-m"' % cmd)
+
+        cmd = 'MASTER_PORT={} GPUS={} GPUS_PER_NODE={} CPUS_PER_TASK={} \
```
ly015, line 78: 1. Please support both `train` and `inference` mode using 'slurm_train.sh` and `slurm_test.py` respectively.

2. Please specify the work dir for each task like `--work-dir {osp.joint(args.root_work_dir, task_name)}`

.dev_scripts/test_benchmark_tmux.py
```diff
@@ -0,0 +1,91 @@
...
+    # ls
+    os.system('tmux send-keys "%s" "C-m"' % 'ls')
+
+    with open(args.config) as config_file:
+        config_list = yaml.load(config_file.read())
+
+    port = args.port
+    window_index = 1
+
+    for model_name, model in config_list.items():
```
ly015, line 62: Using one window for each task may cause trouble if the task number is large. Please split each window into a certain number (e.g. 9, 12 or 16) panes and launch each task in one pane.

For legibility, we can set the pane layout by `tmux select-layout even-vertical` or `tmux select-layout tiled`.

.dev_scripts/benchmark_regression.py
```diff
@@ -0,0 +1,311 @@
...
+
+    parser.add_argument('--gpus', help='the total number of GPUs', default=8)
+    parser.add_argument(
+        '--gpus-per-node',
+        default=8,
+        help='the number of GPUs used per computing node',
+        choices=[1, 2, 3, 4, 5, 6, 7, 8])
+    parser.add_argument(
+        '--cpus-per-task', default=5, help='the number of CPUs used per task')
+    parser.add_argument('--port', default=29666, help='the starting port used')
```
jin-s13, line 48: How about using a random number for port?

.dev_scripts/benchmark_regression_cfg.yaml
```diff
@@ -0,0 +1,143 @@
...
+                cfg-options:
+
+        -   config: configs/body/2d_kpt_sview_rgb_img/topdown_heatmap/mpii/hrnet_w32_mpii_256x256.py
+            priority: 3
+            port: 29675
+            gpus: 8
+            gpus_per_node: 8
+            cpus_per_task: 5
+            partition: openmmlab
+            task_name: train_rnet_w32_mpii_256x256
```
jin-s13, line 140: hrnet?

.dev_scripts/benchmark_regression.py
```diff
@@ -0,0 +1,311 @@
...
+    # select window and pane to run
+    if num_infer_task != 0:
+        num_infer_task_tmp = num_infer_task
+        for i in range(num_infer_win):
+            os.system('tmux select-window -t 0')
+            # tmux select-window -t infer_j
+            os.system(f'tmux select-window -t infer_{i+1}')
+
+            if num_infer_task_tmp >= args.panes_per_window:
+                num_cur_task = args.panes_per_window
```
jin-s13, line 149: What does 'cur' mean?

liqikai9: It means the number of tasks that will be run in the  'current'  window.

.dev_scripts/benchmark_regression_cfg.yaml
liqikai9: typo here

.dev_scripts/benchmark_regression.py
```diff
@@ -0,0 +1,311 @@
...
+
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    args = parse_args()
+
+    if args.root_work_dir is None:
+        date = datetime.datetime.now()
```
ly015, line 58: What about using time.ctime to get the time string?

.dev_scripts/benchmark_regression.py
```diff
@@ -55,251 +74,164 @@ def main():
     args = parse_args()
 
     if args.root_work_dir is None:
-        date = datetime.datetime.now()
-        args.root_work_dir = 'work_dirs/benchmark_regression_' + \
-                             f'{date.year}{date.month:02d}{date.day:02d}_' + \
-                             f'{date.hour:02d}{date.minute:02d}'
-    mmcv.mkdir_or_exist(os.path.abspath(args.root_work_dir))
+        # get the current time stamp
+        ts = time.ctime()
```
ly015, line 78: Consider using `time.strftime` to directly get the time string?

.dev_scripts/benchmark_regression_cfg.yaml
```diff
@@ -1,143 +1,215 @@
...
-            gpus_per_node: 2 # the number of GPUs used each computing node
-            cpus_per_task: 5 # the number of cpus per task
-            partition: openmmlab  # the partition name
-            task_name: infer_hrnet_w32_coco_256x192 # the job name shown in slurm
-            py_args: # other arguments to run the script tools/test.py
-                out: results_hrnet_w32_coco_256x192.json  # output result file
-                eval: mAP # evaluation metric, which depends on the dataset, e.g., "mAP" for MSCOCO
-                fuse-conv-bn:
-                gpu_collect:
+    P0: # the priority of the models, P0: core, P1: important, P2: less important, P3: least important
```
ly015, line 2: A few suggestions about the config file:
- Rename this file "benchmark_regression_cfg_tmpl.yaml" or something, which serves as a template to show the full content that a config file could include. We will add a more compact config file with a full model list and only necessary arguments.
- Use `test` instead of `infer` as the mode name.
- `gpus_per_node` can be set to 8 for all models and modes.

.dev_scripts/benchmark_regression.py
liqikai9: Got it.

.dev_scripts/benchmark_regression_cfg.yaml
liqikai9: Got these suggestions.

jin-s13 (approved):

PR #865, opened 2021-08-11
Update digit_version function

ly015:
Update `digit_version` function according to https://github.com/open-mmlab/mmcv/pull/1185 and https://github.com/open-mmlab/mmcv/pull/1235

See also https://github.com/open-mmlab/mmclassification/pull/402

### DIFF
--- a/mmpose/__init__.py
+++ b/mmpose/__init__.py
@@ -1,19 +1,59 @@
+import warnings
+
 import mmcv
-from mmcv import digit_version, parse_version_info
+from mmcv import parse_version_info
+from packaging.version import parse
 
 from .version import __version__, short_version
 
 mmcv_minimum_version = '1.3.8'
-mmcv_maximum_version = '1.4.0'
+mmcv_maximum_version = '1.5.0'
+
+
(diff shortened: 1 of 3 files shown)


mmpose/__init__.py
```diff
@@ -1,19 +1,59 @@
...
+from packaging.version import parse
 
 from .version import __version__, short_version
 
 mmcv_minimum_version = '1.3.8'
-mmcv_maximum_version = '1.4.0'
+mmcv_maximum_version = '1.5.0'
+
+
+def digit_version(version_str: str, length: int = 4):
```
innerlee, line 13: Importing is good enough, and do not need to copy the implementation.

ly015: You mean importing from mmcv? I think the point of this PR is to avoid import this function from mmcv. Because digit_version from mmcv depends on mmcv version itself, like a circular dependency.

innerlee: Actually my thinking its the reverse. We will use the implementation of digit_version in mmcv, so that we will get automatic bug fixes.

PR #871, opened 2021-08-23
Fix import and deprecation issues in unit tests

zengwang430521:


tests/test_model/test_smpl.py
```diff
@@ -1,10 +1,42 @@
...
 
 from mmpose.models.utils import SMPL
 
 
+def generate_smpl_weight_file(output_dir):
+    """Generate a SMPL model weight file to initialize SMPL model, and generate
+    a 3D joints regressor file."""
+
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
```
innerlee, line 16: Will the generated files be cleaned up after testing?

innerlee (approved):

mmpose/datasets/datasets/body3d/body3d_mpi_inf_3dhp_dataset.py
```diff
@@ -360,7 +360,7 @@ def _report_3d_auc(self, keypoint_results, mode='3dauc'):
 
         preds = np.stack(preds)
         gts = np.stack(gts)
-        masks = np.ones_like(gts[:, :, 0], dtype=np.bool)
+        masks = np.ones_like(gts[:, :, 0], dtype=bool)
```
innerlee, line 363: why these changes?

ly015: The original code raises DeprecationWarning with a reminder that changing like this will not change the behavior.

PR #872, opened 2021-08-24
Add copyright header and pre-commit hook

ly015:
This PR adds a OpenMMLab copyright header to source code files. We also add a pre-commit hook to automatically check and add the copyright header.

See also https://github.com/open-mmlab/mmcv/pull/1181

### Modifications

- Add `.dev_scripts/github/update_copyright.py` as the pre-commit hook entry
- Move 'tools/misc/update_model_index.py` into `.dev_scripts/github/`
- Modify `pre-commit-config.yaml`
- Add copyright headers to relative files.


.dev_scripts/github/update_copyright.py
```diff
@@ -0,0 +1,94 @@
...
+HEADER = 'Copyright (c) OpenMMLab. All rights reserved.\n'
+HEADER_KEYWORDS = {'Copyright', 'License'}
+
+
+def contains_header(lines, comment_symbol):
+    for line in lines:
+        if line.startswith('#!'):
+            # skip shebang line
+            continue
+        elif re.match(f'{comment_symbol}.*({"|".join(HEADER_KEYWORDS)})',
```
innerlee, line 19: In some comments, we might added licence info of reference codes/papers. Will that fail this check?

Maybe checking for the first few (2?) lines is good enough?

demo/mmdetection_cfg/cascade_rcnn_x101_64x4d_fpn_coco.py
```diff
@@ -1,3 +1,4 @@
+# Copyright (c) OpenMMLab. All rights reserved.
```
innerlee, line 1: should configs be ruled out?

ly015: Yes. Will fix it.

.dev_scripts/github/update_copyright.py
ly015: Ok

ly015: fixed.

demo/mmdetection_cfg/cascade_rcnn_x101_64x4d_fpn_coco.py
ly015: Fixed by adding "exclude" to `pre-commit-config.yaml`

PR #873, opened 2021-08-25
Correct dataset preparation guide of WFLW

zhiqwang:
Seems that the annotation files is redundant, so I'm removing them.


jin-s13 (approved):

jin-s13:
Thanks!

PR #905, opened 2021-09-06
Hourglass ae net

luminxu:
Add AE version Hourglass backbone to reimplement the paper. We add the backbone and conduct corresponding modification.


jin-s13:
Fix some conflicts before merging.

PR #932, opened 2021-09-22
[Feature] Add PoseWarper detector model

liqikai9:
## Motivation

To support input multiple frames as a single sample when dealing with video dataset, we add a new detector `PoseWarper` which is inherited from `TopDown`. The main difference between `PoseWarper` and `TopDown` lies in that the former takes a list of tensors as input while the latter takes a single tensor as input.

To support [PoseWarper](https://arxiv.org/abs/1906.04016), we add a new neck `PoseWarperNeck` in `mmpose/models/necks/posewarper_neck.py`.

## Modification

1. Add `mmpose/models/detectors/posewarper.py`.
Modify `mmpose/models/detectors/__init__.py`

2. Add `mmpose/models/necks/posewarper_neck.py`.
Modify `mmpose/models/necks/__init__.py`.

3. To support freezing certain stages in the backbone `HRNet`, also modify `mmpose/models/backbones/hrnet.py`.
Modify  `mmpose/models/detectors/top_down.py`, fix typo.

4. Add related unittests in `tests/test_backbones/test_hrnet.py`, `tests/test_models/test_top_down_forward.py` and `tests/test_necks/test_posewarper_neck.py`.

5. To support more flexible input format, modify `tools/analysis/get_flops.py`.


mmpose/models/necks/posewarper_neck.py
```diff
@@ -0,0 +1,268 @@
...
+                offset = (self.offset_layers[j](diff_feature))
+                warped_heatmap_tmp = self.deform_conv_layers[j](ref_feature,
+                                                                offset)
+                warped_heatmap += warped_heatmap_tmp / num_offset_layers
+
+            return warped_heatmap
+
+        else:  # test mode, multi frames
+            if num_frames == 5:
+                weight_frame = [0.3, 0.25, 0.25, 0.1, 0.1]
```
jin-s13, line 242: We should not hard-code these here.

jin-s13: Move it to cfg

mmpose/models/necks/posewarper_neck.py
```diff
@@ -0,0 +1,268 @@
...
+                weight_frame = [0.4, 0.3, 0.3]
+            elif num_frames == 7:
+                weight_frame = [0.3, 0.2, 0.2, 0.1, 0.1, 0.05, 0.05]
+            elif num_frames % 2 == 0:
+                raise ValueError('Number of frames must be odd number')
+            else:
+                weight_frame = [1.0 / num_frames for _ in range(num_frames)]
+
+            diff_features = []
+            for i in range(num_frames):
```
luminxu, line 253: start from 1?

mmpose/models/necks/posewarper_neck.py
```diff
@@ -0,0 +1,268 @@
...
+        num_offset_layers = len(self.offset_layers)
+
+        # batch_size, _, heatmap_height, heatmap_width = inputs[0].shape
+        for i in range(num_frames):
+            inputs[i] = self._transform_inputs(inputs[i])
+            inputs[i] = self.final_layer(inputs[i])
+
+        ref_feature = inputs[0]
+        # calculate difference features
+        if num_frames == 2:  # train mode, input two frames
```
luminxu, line 227: I think num_frames == 2 is not a very special case.

luminxu (commented):
1. According to our previous discussion, maybe we first merge the individual modules (dataset, detector, etc) into a new branch (e.g. posewarper). After the whole algorithm is finished and verified, we merge it to the master.
2. Maybe we should rename the detector as posewarper.py or something else, as this detector does not work for all the top-down video pose estimation methods. For example, some top-down video methods merge features in another way, or apply different backbones for different frames.

luminxu (approved):

mmpose/models/detectors/posewarper.py
```diff
@@ -0,0 +1,235 @@
...
+                - "scale": scale of the bbox
+                - "rotation": rotation of the bbox
+                - "bbox_score": score of bbox
+            return_loss (bool): Option to `return loss`. `return loss=True`
+                for training, `return loss=False` for validation & test.
+            return_heatmap (bool) : Option to return heatmap.
+
+        Returns:
+            dict|tuple: if `return loss` is true, then return losses.
+              Otherwise, return predicted poses, boxes, image paths
```
jin-s13, line 98: The indent looks strange here.

mmpose/models/detectors/top_down.py
```diff
@@ -293,8 +292,9 @@ def show_result(self,
                 font_scale=font_scale,
                 show=False)
 
-        imshow_keypoints(img, pose_result, skeleton, kpt_score_thr,
-                         pose_kpt_color, pose_link_color, radius, thickness)
+            imshow_keypoints(img, pose_result, skeleton, kpt_score_thr,
```
jin-s13, line 295: Please double check this change.

mmpose/models/detectors/posewarper.py
liqikai9: I will fix the indent in detector `posewarper` and `top_down`.

mmpose/models/detectors/top_down.py
liqikai9: I will check this.

jin-s13 (approved):

mmpose/models/detectors/posewarper.py
```diff
@@ -0,0 +1,235 @@
...
+    from mmcv.runner import auto_fp16
+except ImportError:
+    warnings.warn('auto_fp16 from mmpose will be deprecated from v0.15.0'
+                  'Please install mmcv>=1.1.4')
+    from mmpose.core import auto_fp16
+
+
+@POSENETS.register_module()
+class PoseWarper(TopDown):
+    """Top-down pose detectors for video inputs.
```
ly015, line 20: If this class is just for the pose warper algorithm, instead of general video pose estimation, the description here should be more specific, and include paper reference.

liqikai9: OK, I will change it.

mmpose/models/necks/posewarper_neck.py
```diff
@@ -0,0 +1,311 @@
...
+                 extra=None,
+                 res_blocks={},
+                 offsets={},
+                 deform_conv={},
+                 in_index=0,
+                 input_transform=None,
+                 freeze_trans_layer=False,
+                 norm_eval=False,
+                 im2col_step=80,
+                 least_mmcv_version='1.3.16'):
```
ly015, line 68: `least_mmcv_version` should not be an argument. Maybe make it a class attribute, or just hard code it in the version check. And I suggest using the name `minimum_mmcv_version`.

mmpose/models/necks/posewarper_neck.py
```diff
@@ -0,0 +1,311 @@
...
+    blocks_dict = {'BASIC': BasicBlock, 'BOTTLENECK': Bottleneck}
+
+    def __init__(self,
+                 in_channels,
+                 out_channels,
+                 inner_channels,
+                 deform_groups=17,
+                 dilations=(3, 6, 12, 18, 24),
+                 extra=None,
+                 res_blocks={},
```
ly015, line 60: The arguments `res_blocks`, `offsets` and `deform_conv` may need better design:
1. We should avoid using mutable datatype like dict as the default argument value. 
2. The docstring provides no information about what should be contained in these arguments. This could be confusing to users.
3. Use more specific names like `res_blocks_cfg`.

mmpose/models/necks/posewarper_neck.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    blocks_dict = {'BASIC': BasicBlock, 'BOTTLENECK': Bottleneck}
+
+    def __init__(self,
+                 in_channels,
+                 out_channels,
+                 inner_channels,
+                 deform_groups=17,
+                 dilations=(3, 6, 12, 18, 24),
+                 extra=None,
```
ly015, line 59: The name `extra` is ambiguous. Please use a more specific name.

ly015:
And please solve the conflict by rebasing onto dev-0.21. @liqikai9

mmpose/models/necks/posewarper_neck.py
liqikai9: OK, I will make it a class attribute and use the name `minimum_mmcv_version` instead.

mmpose/models/necks/posewarper_neck.py
liqikai9: OK, I will use the name `res_blocks_cfg` instead. Since the `offsets` and `deform_conv` only have one argument `kernel` in the dict,  to be more specific, I use the name `offsets_kernel` and `deform_conv_kernel` instead, and change their type to `int`.

mmpose/models/necks/posewarper_neck.py
liqikai9: OK, since  `extra` only have one effective argument `trans_conv_kernel`, I will change `extra` into `trans_conv_kernel` instead, and the type is `int`.

PR #1098, opened 2021-12-23
[WebcamAPI] Supersaiyan

jin-s13:
## Checklist

**Before PR**:

**After PR**:


mmpose/core/visualization/effects.py
```diff
@@ -3,6 +3,45 @@
 import numpy as np
 
 
+def screen_matting(img, color_low=None, color_high=None, color=None):
```
ly015, line 6: What about moving these codes into tools/webcam/. For example in the node.py or utils.py.

tools/webcam/webcam_apis/nodes/frame_effect_nodes.py
```diff
@@ -20,6 +23,44 @@
     psutil_proc = None
 
 
+def is_url(filename):
```
ly015, line 26: Maybe move it into webam_apis/utils?

jin-s13: OK.

mmpose/core/visualization/effects.py
jin-s13: Done.

PR #1215, opened 2022-03-03
[Feature] Add PoseFormer backbone

QwQ2000:
## Modification

Add backbone, head and config of the PoseFormer (ICCV 2021) into the repository.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/models/backbones/poseformer.py
+++ b/mmpose/models/backbones/poseformer.py
@@ -1,0 +1,315 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import torch
+import torch.nn as nn
+from einops import rearrange
+from mmcv.cnn import build_norm_layer
+from mmcv.cnn.bricks.registry import DROPOUT_LAYERS
+from mmcv.cnn.bricks.transformer import FFN
+from mmcv.runner.base_module import BaseModule
+
+from ..builder import BACKBONES
+from .base_backbone import BaseBackbone
+
+
+class MultiheadAttention(BaseModule):
+    """Multi-head Attention Module.
+
+    This module implements multi-head attention that supports different input
+    dims and embed dims. And it also supports a shortcut from ``value``, which
+    is useful if input dims is not the same with embed dims.
+
+    Args:
+        embed_dims (int): The embedding dimension.
+        num_heads (int): Parallel attention heads.
+        input_dims (int, optional): The input dimension, and if None,
+            use ``embed_dims``. Defaults to None.
+        attn_drop (float): Dropout rate of the dropout layer after the
+            attention calculation of query and key. Defaults to 0.
+        proj_drop (float): Dropout rate of the dropout layer after the
+            output projection. Defaults to 0.
+        dropout_layer (dict): The dropout config before adding the shortcut.
+            Defaults to ``dict(type='Dropout', drop_prob=0.)``.
+        qkv_bias (bool): If True, add a learnable bias to q, k, v.
+            Defaults to True.
+        qk_scale (float, optional): Override default qk scale of
+            ``head_dim ** -0.5`` if set. Defaults to None.
+        proj_bias (bool) If True, add a learnable bias to output projection.
+            Defaults to True.
+        v_shortcut (bool): Add a shortcut from value to output. It's usually
+            used if ``input_dims`` is different from ``embed_dims``.
+            Defaults to False.
+        init_cfg (dict, optional): The Config for initialization.
+            Defaults to None.
+    """
+
+    def __init__(self,
+                 embed_dims,
+                 num_heads,
+                 input_dims=None,
+                 attn_drop=0.,
+                 proj_drop=0.,
+                 dropout_layer=dict(type='Dropout', drop_prob=0.),
+                 qkv_bias=True,
+                 qk_scale=None,
+                 proj_bias=True,
+                 v_shortcut=False,
+                 init_cfg=None):
+        super(MultiheadAttention, self).__init__(init_cfg=init_cfg)
+
+        self.input_dims = input_dims or embed_dims
+        self.embed_dims = embed_dims
+        self.num_heads = num_heads
+        self.v_shortcut = v_shortcut
+
+        self.head_dims = embed_dims // num_heads
+        self.scale = qk_scale or self.head_dims**-0.5
+
+        self.qkv = nn.Linear(self.input_dims, embed_dims * 3, bias=qkv_bias)
+        self.attn_drop = nn.Dropout(attn_drop)
+        self.proj = nn.Linear(embed_dims, embed_dims, bias=proj_bias)
+        self.proj_drop = nn.Dropout(proj_drop)
+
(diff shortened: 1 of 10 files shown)


mmpose/models/backbones/poseformer.py
```diff
@@ -0,0 +1,312 @@
...
+from mmcv.cnn import build_norm_layer
+from mmcv.cnn.bricks.registry import DROPOUT_LAYERS
+from mmcv.cnn.bricks.transformer import FFN
+from mmcv.runner.base_module import BaseModule
+
+from ..builder import BACKBONES
+from .base_backbone import BaseBackbone
+
+
+class MultiheadAttention(BaseModule):
```
jin-s13, line 13: mmcv also has MultiheadAttention module. Can we use that one?

mmpose/models/backbones/poseformer.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+@BACKBONES.register_module()
+class PoseFormer(BaseBackbone):
+
+    def __init__(self,
+                 num_frame=9,
+                 num_joints=17,
+                 in_chans=2,
+                 embed_dim_ratio=32,
```
liqikai9, line 185: Does `embed_dim_ratio`  mean the embedding dimension for each joint? How about using `embed_dim_per_joint` ?

mmpose/models/backbones/poseformer.py
```diff
@@ -0,0 +1,312 @@
...
+                 depth=4,
```
liqikai9, line 186: Is the `depth` for spatial and temporal transformer always the same? If not, do we need to distinguish them?

liqikai9, line 186: Is the `depth` for spatial and temporal transformer always the same? If not, do we need to distinguish them?

mmpose/models/backbones/poseformer.py
```diff
@@ -0,0 +1,312 @@
...
+            attn_drop_rate (float): attention dropout rate
+            drop_path_rate (float): stochastic depth rate
+            norm_layer: (nn.Module): normalization layer
+        """
+        super().__init__()
+
+        if norm_cfg is None:
+            norm_cfg = dict(type='LN')
+        # Temporal embed_dim is num_joints * spatial embedding dim ratio
+        embed_dim = embed_dim_ratio * num_joints
```
liqikai9, line 220: Does `embed_dim ` mean the embedding dimension for each frame? How about using `embed_dim_per_frame `?

liqikai9, line 220: Does `embed_dim ` mean the embedding dimension for each frame? How about using `embed_dim_per_frame `?

mmpose/models/backbones/poseformer.py
```diff
@@ -0,0 +1,312 @@
...
+                qk_scale=qk_scale,
+                drop_rate=drop_rate,
+                attn_drop_rate=attn_drop_rate,
+                drop_path_rate=dpr[i],
+                act_cfg=act_cfg,
+                norm_cfg=norm_cfg,
+                init_cfg=init_cfg) for i in range(depth)
+        ])
+
+        self.blocks = nn.ModuleList([
```
liqikai9, line 250: Does `self.blocks` mean the temporal transformer blocks? How about using `self.temporal_blocks` for clarity?

liqikai9, line 250: Does `self.blocks` mean the temporal transformer blocks? How about using `self.temporal _blocks` for clarity?

mmpose/models/backbones/poseformer.py
```diff
@@ -0,0 +1,312 @@
...
+        x = (attn @ v).transpose(1, 2).reshape(B, N, self.embed_dims)
+        x = self.proj(x)
+        x = self.out_drop(self.proj_drop(x))
+
+        if self.v_shortcut:
+            x = v.squeeze(1) + x
+        return x
+
+
+class TransformerEncoderLayer(BaseModule):
```
liqikai9, line 92: mmcv has the similar layer `BaseTransformerLayer`module under `mmcv/cnn/bricks/transformer.py`, which also includes `MultiheadAttention`. Can we use that one?

liqikai9, line 92: mmcv has the similar layer `BaseTransformerLayer`module under `mmcv/cnn/bricks/transformer.py`, which also includes `MultiheadAttention`. Can we use that one?

mmpose/models/backbones/poseformer.py
QwQ2000: This MultiheadAttention module is adapted from **mmcls**. Compared with the MultiheadAttention in **mmcv**, its counterpart in **mmcls** is more similar to the implementation in the official **PoseFormer** repository.

mmpose/models/backbones/poseformer.py
QwQ2000: This TransformerEncoderLayer module is adapted from **mmcls**. Compared with the BaseTransformerLayer in **mmcv**, its counterpart in **mmcls** is more similar to the implementation in the official PoseFormer repository.

mmpose/models/backbones/poseformer.py
QwQ2000: I will give **mmcv** Transformer blocks a try. However, if the output and the model weights of the re-implemented **PoseFormer** based on **mmcv** Transformer blocks could not fit the official version of **PoseFormer**, keeping the **mmcls** Transformer modules might be a good choice.

mmpose/models/backbones/poseformer.py
QwQ2000: I will give **mmcv** Transformer blocks a try. However, if the output and the model weights of the re-implemented **PoseFormer** based on **mmcv** Transformer blocks could not fit the official version of **PoseFormer**, keeping the **mmcls** Transformer modules might be a good choice.

mmpose/models/backbones/poseformer.py
QwQ2000: In the official implementation of **PoseFormer**, the spatial and temporal transformer share the same depth. 
However, I think it's obviously better to distinguish the two depths for clarity and I will take this comment.

mmpose/models/backbones/poseformer.py
QwQ2000: **self.temporal_blocks** is a better choice.

mmpose/models/backbones/poseformer.py
QwQ2000: Now I use **spatial_embed_dim** and **temporal_embed_dim** instead of **embed_dim_ratio** and **embed_dim**.

darcula1993:
what's wrong with this branch? Is ther any update?
@QwQ2000 @ly015

QwQ2000:
Sorry, I have resigned from the Shanghai AI Lab for a long time. I'm busy working on other projects so I don't have much time for this. If someone could help me finish this feature, I will be always willing to help.

PR #1234, opened 2022-03-10
[Fix] fix unmatched MMCV version in DockerFile 

### LINKED ISSUE #1232: The docker image built by Dockerfille, when I import mmpose, there is an error that mmcv version does not match
Thanks for your error report and we appreciate it a lot.
If you feel we have helped you, give us a STAR! :satisfied:

**Checklist**

1. I have searched related issues but cannot get the expected help.
1. The bug has not been fixed in the latest version.

**Describe the bug**

**Reproduction**

- What command or script did you run?

```
A placeholder for the command.
```

- What config did you run?

```
A placeholder for the config.
```

- Did you make any modifications on the code or config? Did you understand what you have modified?
- What dataset did you use?

**Environment**

1. Please run `PYTHONPATH=${PWD}:$PYTHONPATH python mmpose/utils/collect_env.py` to collect necessary environment information and paste it here.
1. You may add addition that may be helpful for locating the problem, such as
  - How you installed PyTorch [e.g., pip, conda, source]
  - Other environment variables that may be related (such as `$PATH`, `$LD_LIBRARY_PATH`, `$PYTHONPATH`, etc.)

**Error traceback**

If applicable, paste the error traceback here.

```
A placeholder for traceback.
```

**Bug fix**

If you have already identified the reason, you can provide the information here. If you are willing to create a PR to fix it, please also leave a comment here and that would be much appreciated!

raykindle:
## Motivation

Resolve #1232 

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,7 +17,8 @@ RUN pip install cython
 RUN pip install xtcocotools
 
 # Install MMCV
-RUN pip install mmcv-full==latest+torch1.6.0+cu101 -f https://download.openmmlab.com/mmcv/dist/index.html
+RUN pip install --no-cache-dir --upgrade pip wheel setuptools
+RUN pip install --no-cache-dir mmcv-full==1.3.17 -f https://download.openmmlab.com/mmcv/dist/cu101/torch1.6.0/index.html
 
 # Install MMPose
 RUN conda clean --all


ly015:
@raykindle Thank you very much for your contribution. Could you please sign the CLA？

docker/Dockerfile
```diff
@@ -17,7 +17,8 @@ RUN pip install cython
 RUN pip install xtcocotools
 
 # Install MMCV
-RUN pip install mmcv-full==latest+torch1.6.0+cu101 -f https://download.openmmlab.com/mmcv/dist/index.html
+RUN pip install --no-cache-dir --upgrade pip wheel setuptools
+RUN pip install --no-cache-dir mmcv-full==1.3.17 -f https://download.openmmlab.com/mmcv/dist/cu101/torch1.6.0/index.html
```
ly015, line 21: What about using `-U mmcv-full` instead of specifying the version?

raykindle: It may report an error，because downstream libraries may have version restrictions。
e.g.：
when I import mmpose, there appears "AssertionError: MMCV==1.3.5 is used but incompatible. Please install mmcv>=1.3.8, <=1.5.0"

ly015: Sure, we can merge this pr as it is now.  We will check if there is a better way to install the latest mmcv in docker in the future. Thank you.

ly015 (approved):

PR #1242, opened 2022-03-15
[Fix] Changed channel size from -1 to an operable number "channel_shuffle"

PINTO0309:
## Motivation
If we use **`pytorch2onnx.py`** to generate ONNX with the batch size set to a variable size of **`-1`**, it will generate a model that cannot be inferred by onnxruntime because the **`Reshape`** operation on ONNX will include two **`-1`** in the modified shape. **`(1)`** in the figure below. The ONNX model shown below will cause an error in onnxruntime if the batch size is changed to **`-1`**. Because onnxruntime does not allow more than two dimensions of variable size.

- Sample ONNX file just before initializing batch size
image
[litehrnet_18_coco_Nx256x192_reshape_channel=-1.onnx.zip](https://github.com/open-mmlab/mmpose/files/8254335/litehrnet_18_coco_Nx256x192_reshape_channel.-1.onnx.zip)

The ONNX model output after applying this modification is shown below. The batch size is initialized to **`(1) -1`**, but the channel locations are replaced with fixed values **`(2) 32`** computed, so that onnxruntime will not generate errors and can be inferred correctly.
- Sample ONNX file output immediately after initializing batch size
image
[litehrnet_18_coco_Nx256x192.onnx.zip](https://github.com/open-mmlab/mmpose/files/8254343/litehrnet_18_coco_Nx256x192.onnx.zip)

## Modification
Replace the channels part with the value calculated by **`groups * channels_per_group`** instead of **`-1`** so that onnxruntime can infer correctly when the batch size is changed to variable.

From:
```python
assert (num_channels % groups == 0), ('num_channels should be '
                                      'divisible by groups')
channels_per_group = num_channels // groups

x = x.view(batch_size, groups, channels_per_group, height, width)
x = torch.transpose(x, 1, 2).contiguous()
x = x.view(batch_size, -1, height, width)
```

To:
```python
assert (num_channels % groups == 0), ('num_channels should be '
                                      'divisible by groups')
channels_per_group = num_channels // groups

x = x.view(batch_size, groups, channels_per_group, height, width)
x = torch.transpose(x, 1, 2).contiguous()
x = x.view(batch_size, groups * channels_per_group, height, width)
```

## BC-breaking (Optional)
None in particular.

## Use cases (Optional)
None in particular.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/models/backbones/utils/channel_shuffle.py
+++ b/mmpose/models/backbones/utils/channel_shuffle.py
@@ -24,6 +24,6 @@ def channel_shuffle(x, groups):
 
     x = x.view(batch_size, groups, channels_per_group, height, width)
     x = torch.transpose(x, 1, 2).contiguous()
-    x = x.view(batch_size, -1, height, width)
+    x = x.view(batch_size, groups * channels_per_group, height, width)
 
     return x


jin-s13 (approved):

ly015:
@PINTO0309 Thank you very much for contributing to MMPose.

Also please note that the deployment of MMPose models has been supported in [MMDeploy](https://github.com/open-mmlab/mmdeploy), the OpenMMLab model deployment framework. Consider giving it a try :)

Issue #1261, opened 2022-03-25
2d hand keypoint order problem of RHD dataset


tiger990111:
I am now trying to finetune 2d hand pose model on coco_hand, freihand, rhd. But I find that the keypoint order of rhd is different.

here is my finetune config setting of dataset.

```
_base_ = [
    '../../../../_base_/default_runtime.py',
    '../../../../_base_/datasets/coco_wholebody_hand.py'
]

data_root = [
             '/data/pose/hand2d/coco',
             '/data/pose/hand2d/rhd',
             '/data/pose/hand2d/freihand'
             ]
data = dict(
    samples_per_gpu=64,# Batch size of each single GPU during training
    workers_per_gpu=8, # Workers to pre-fetch data for each single GPU
    val_dataloader=dict(samples_per_gpu=32),# Batch size of each single GPU during validation
    test_dataloader=dict(samples_per_gpu=32),
    train=[
        dict(
            type='HandCocoWholeBodyDataset',  # Name of dataset
            ann_file=f'{data_root[0]}/annotations/coco_wholebody_train_v1.0.json',
            img_prefix=f'{data_root[0]}/train2017/',
            data_cfg=data_cfg,
            pipeline=train_pipeline,
            dataset_info={{_base_.dataset_info}}
        ),

        dict(
            type='Rhd2DDataset',
            ann_file=f'{data_root[1]}/annotations/rhd_train.json',
            img_prefix=f'{data_root[1]}/',
            data_cfg=data_cfg,
            pipeline=train_pipeline,
            dataset_info={{_base_.dataset_info}}),

        dict(
            type='FreiHandDataset',
            ann_file=f'{data_root[2]}/annotations/freihand_train.json',
            img_prefix=f'{data_root[2]}/',
            data_cfg=data_cfg,
            pipeline=train_pipeline,
            dataset_info={{_base_.dataset_info}})
    ],

    val=dict(
        type='HandCocoWholeBodyDataset',
        ann_file=f'{data_root[0]}/annotations/coco_wholebody_val_v1.0.json',
        img_prefix=f'{data_root[0]}/val2017/',
        data_cfg=data_cfg,
        pipeline=val_pipeline,
        dataset_info={{_base_.dataset_info}}
        ),
    test=dict(
        type='HandCocoWholeBodyDataset',
        ann_file=f'{data_root[0]}/annotations/coco_wholebody_val_v1.0.json',
        img_prefix=f'{data_root[0]}/val2017/',
[...]

PR #1265, opened 2022-03-28
[Fix] Fix `keypoint_info` of RHD dataset 

liqikai9:
## Motivation

Fix issue #1261.

## Modification

Fix the `skeleton_info` of RHD dataset info under  `mmpose/configs/_base_/datasets/rhd2d.py` by these modifications:

1.`link=('wrist', 'thumb1') `->` link=('wrist', 'thumb4')`

2.`link=('wrist', 'forefinger1')` -> `link=('wrist', 'forefinger4')`

3.l`ink=('wrist', 'middle_finger1') `-> `link=('wrist', 'middle_finger4')`

4.`link=('wrist', 'ring_finger1') `-> `link=('wrist', 'ring_finger4')`

5.`link=('wrist', 'pinky_finger1')` -> `link=('wrist', 'pinky_finger4')`

These modifications are minor, which will not affect the trained models.

Further, fix the paper info of COCO-Wholebody-Hand dataset under `mmpose/docs/en/tasks/2d_hand_keypoint.md`

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/datasets/datasets/hand/rhd2d_dataset.py
+++ b/mmpose/datasets/datasets/hand/rhd2d_dataset.py
@@ -26,26 +26,26 @@ class Rhd2DDataset(Kpt2dSviewRgbImgTopDownDataset):
     Rhd keypoint indexes::
 
         0: 'wrist',
-        1: 'thumb1',
-        2: 'thumb2',
-        3: 'thumb3',
-        4: 'thumb4',
-        5: 'forefinger1',
-        6: 'forefinger2',
-        7: 'forefinger3',
-        8: 'forefinger4',
-        9: 'middle_finger1',
-        10: 'middle_finger2',
-        11: 'middle_finger3',
-        12: 'middle_finger4',
-        13: 'ring_finger1',
-        14: 'ring_finger2',
-        15: 'ring_finger3',
-        16: 'ring_finger4',
-        17: 'pinky_finger1',
-        18: 'pinky_finger2',
-        19: 'pinky_finger3',
-        20: 'pinky_finger4'
+        1: 'thumb4',
+        2: 'thumb3',
+        3: 'thumb2',
+        4: 'thumb1',
+        5: 'forefinger4',
+        6: 'forefinger3',
+        7: 'forefinger2',
+        8: 'forefinger1',
+        9: 'middle_finger4',
+        10: 'middle_finger3',
+        11: 'middle_finger2',
+        12: 'middle_finger1',
+        13: 'ring_finger4',
+        14: 'ring_finger3',
+        15: 'ring_finger2',
+        16: 'ring_finger1',
+        17: 'pinky_finger4',
+        18: 'pinky_finger3',
+        19: 'pinky_finger2',
+        20: 'pinky_finger1'
 
     Args:
         ann_file (str): Path to the annotation file.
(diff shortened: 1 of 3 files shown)


ly015:
@liqikai9 Could you please double-check the keypoint order by visualizing them with the demo script?

jin-s13:
This fix will cause some undesirable confusion. Since in other hand datasets, 'wrist' is connected to 'finger1' (this keypoint has its unique semantic meaning). 

Maybe we should take a look at the definition of RHD hand.

liqikai9:
> @liqikai9 Could you please double-check the keypoint order by visualizing them with the demo script?

Yes, I have checked the demo script using these commands:

`python demo/top_down_img_demo.py configs/hand/2d_kpt_sview_rgb_img/topdown_heatmap/rhd2d/res50_rhd2d_256x256.py checkpoints/res50_rhd2d_256x256-5dc7e4cc_20210330.pth --img-root tests/data/rhd/ --json-file tests/data/rhd/test_rhd.json --out-img-root vis_results_th_rhd`

and 

`python demo/top_down_img_demo.py configs/hand/2d_kpt_sview_rgb_img/deeppose/rhd2d/res50_rhd2d_256x256.py checkpoints/deeppose_res50_rhd2d_256x256-37f1c4d3_20210330.pth --img-root tests/data/rhd/ --json-file tests/data/rhd/test_rhd.json --out-img-root vis_results_dp_rhd`.

And here are the visualization results, respectively.

topdown_heatmap_rhd_vis_0   deeppose_rhd_vis_0

The visualization results look just fine.

jin-s13:
Many users would like to combine different hand datasets to train a single model. Since the keypoint definition is different, it will cause some problems.

liqikai9:
> 

@jin-s13 Yes, itis a little bit confused. Here is the definition of keypoint indices of [RHD dataset](https://lmb.informatik.uni-freiburg.de/resources/datasets/RenderedHandpose/README). 
It says, `1-4: left thumb [tip to palm]`, in which the finger is from tip to palm, so I think the `wrist` should be connected be `finger4`. 

Maybe it is different from the other hand datasets like freihand and coco_wholebody_hand.

liqikai9:
> Many users would like to combine different hand datasets to train a single model. Since the keypoint definition is different, it will cause some problems.

@jin-s13  That's right. But if we change the keypoint indices of these fingers, we may need to train the models related to RHD dataset again, because the order of final channels corresponding to different keypoint dices is different. 

So, do you think it's a good idea to unify these keypoint definitions and train the models again (or add extra process during inference), or can we just change the skeleton connection?

jin-s13:
I think we should add a **warning notice** in RHD dataset. Its keypoint definition is different from others.
Also it is a good idea to rename the keypoint. 1-4 is confusing.

mmpose/datasets/datasets/hand/rhd2d_dataset.py
```diff
@@ -73,6 +73,11 @@ def __init__(self,
                 'for details.', DeprecationWarning)
             cfg = Config.fromfile('configs/_base_/datasets/rhd2d.py')
             dataset_info = cfg._cfg_dict['dataset_info']
+        warnings.warn(
+            'Please note that the in RHD dataset, its keypoint definition is'
+            'different from others. Check `configs/_base_/datasets/rhd2d.py`'
```
ly015, line 78: 'others' here is ambiguous. Please list some specific dataset names like coco.

jin-s13 (approved):

PR #1279, opened 2022-03-31
[Feature] Support SmoothNet 

ailingzengzzz:
## Motivation

Support SmoothNet for pose smoothing

## Modification

- Add `SmoothNet` and `SmoothNetFilter`
- Support filter shared by multiple targets
- Update docs

## BC-breaking (Optional)

-

## Use cases (Optional)

-

## Checklist

**Before PR**:

**After PR**:


README.md
```diff
@@ -137,6 +137,7 @@ A summary can be found in the [Model Zoo](https://mmpose.readthedocs.io/en/lates
 * [x] [UDP](https://mmpose.readthedocs.io/en/latest/papers/techniques.html#udp-cvpr-2020) (CVPR'2020)
 * [x] [Albumentations](https://mmpose.readthedocs.io/en/latest/papers/techniques.html#albumentations-information-2020) (Information'2020)
 * [x] [SoftWingloss](https://mmpose.readthedocs.io/en/latest/papers/techniques.html#softwingloss-tip-2021) (TIP'2021)
+* [x] [SmoothNet](/configs/_base_/filters/smoothnet_h36m.md) (arXiv'2021)
```
jin-s13, line 140: The link looks different from others. Shall we use something that starts with `https://`？

ly015: We don't have a readthedocs page or link for SmoothNet.

This PR only supports using SmoothNet as a temporal filter. The model training and testing are not included, thus there's no corresponding config files. And SmoothNet doesn't belong to any tasks we currently have, so the .md file is placed in `configs/_base_` and can't be collected when building the doc pages.

So I think we can link to the .md file here for now. And

ly015: We can add the readthedocs page for SmoothNet when the training is supported. Or we can reorganize the model zoo page to add something like 'plugin' as a new category.

ly015 (approved):

README.md
jin-s13: understood.

jin-s13 (approved):

PR #1320, opened 2022-04-15
[Enhance] Speed up the inference by reducing numpy array copy and doing Normalization in GPU

### LINKED ISSUE #1316: Speedup the 'TopDownAffine' by give short cut to rotation=0
The `TopDownAffine` is a **disaster** for cpu load. Because in most case of inference, there is no ROTATION at all. Please give a shortcut when rotation==0. The cv2.warpAffine is so laborious.
```
# mmpose/datasets/pipelines/top_down_transform.py
@PIPELINES.register_module()
class TopDownAffine:
      ...
        c = results['center']
        s = results['scale']
        r = results['rotation']
         ....
        img = [
                    cv2.warpAffine(
                        i,
                        trans, (int(image_size[0]), int(image_size[1])),
                        flags=cv2.INTER_LINEAR) for i in img
                ]
```

the configure file 
```
#configs/wholebody/2d_kpt_sview_rgb_img/topdown_heatmap/coco-wholebody/res152_coco_wholebody_384x288.py
val_pipeline = [
    dict(type='LoadImageFromFile'),
    dict(type='TopDownAffine'),
    dict(type='ToTensor'),
    dict(
        type='NormalizeTensor',
        mean=[0.485, 0.456, 0.406],
        std=[0.229, 0.224, 0.225]),
    dict(
        type='Collect',
        keys=['img'],
        meta_keys=[
            'image_file', 'center', 'scale', 'rotation', 'bbox_score',
            'flip_pairs'
        ]),
]
```
I used GPU to do inference of a video by `demo/top_down_video_demo_full_frame_without_det.py`. However, the cpu workload is surprising 2000%. Part of the result own to that I used 1280x800 resolution video and 512x512 feature size, but the most because of this laborious `TopDownAffine`.

chenxinfeng4:
Allow ToTensor to automaticly upload data to gpu, so that Normalize will be done in GPU.
Fix #1316

## Motivation

The inference for big image size is so slow. Because there are so many numpy array coping. And the data normalize should also be done in GPU if possible. After my revision, the cpu usage is significantly reduced, and the speed is increased.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/datasets/pipelines/shared_transform.py
+++ b/mmpose/datasets/pipelines/shared_transform.py
@@ -4,6 +4,7 @@ from collections.abc import Sequence
 
 import mmcv
 import numpy as np
+import torch
 from mmcv.parallel import DataContainer as DC
 from mmcv.utils import build_from_cfg
 from numpy import random
@@ -27,11 +28,18 @@ class ToTensor:
         results (dict): contain all information about training.
     """
 
+    def __init__(self, device='cpu'):
+        self.device = device
+
     def __call__(self, results):
+        to_tensor = (lambda x: torch.from_numpy(x.astype('float32')).permute(
+            2, 0, 1).to(self.device))
         if isinstance(results['img'], (list, tuple)):
-            results['img'] = [F.to_tensor(img) for img in results['img']]
+            results['img'] = [
+                to_tensor(img.astype('float32')) for img in results['img']
+            ]
         else:
-            results['img'] = F.to_tensor(results['img'])
+            results['img'] = to_tensor(results['img'].astype('float32'))
 
         return results
 
@@ -54,12 +62,12 @@ class NormalizeTensor:
     def __call__(self, results):
         if isinstance(results['img'], (list, tuple)):
             results['img'] = [
-                F.normalize(img, mean=self.mean, std=self.std)
+                F.normalize(img, mean=self.mean, std=self.std, inplace=True)
                 for img in results['img']
             ]
         else:
             results['img'] = F.normalize(
-                results['img'], mean=self.mean, std=self.std)
+                results['img'], mean=self.mean, std=self.std, inplace=True)
 
         return results
 
--- a/mmpose/datasets/pipelines/shared_transform.py
+++ b/mmpose/datasets/pipelines/shared_transform.py
@@ -35,9 +35,7 @@ class ToTensor:
         to_tensor = (lambda x: torch.from_numpy(x.astype('float32')).permute(
             2, 0, 1).to(self.device))
         if isinstance(results['img'], (list, tuple)):
-            results['img'] = [
-                to_tensor(img) for img in results['img']
-            ]
+            results['img'] = [to_tensor(img) for img in results['img']]
         else:
             results['img'] = to_tensor(results['img'])
 
--- a/mmpose/datasets/pipelines/shared_transform.py
+++ b/mmpose/datasets/pipelines/shared_transform.py
@@ -31,13 +31,15 @@ class ToTensor:
     def __init__(self, device='cpu'):
         self.device = device
 
+    def _to_tensor(self, x):
+        return torch.from_numpy(x.astype('float32')).permute(2, 0, 1).to(
+            self.device).div_(255.0)
+
     def __call__(self, results):
-        to_tensor = (lambda x: torch.from_numpy(x.astype('float32')).permute(
-            2, 0, 1).to(self.device) / 255.0)
         if isinstance(results['img'], (list, tuple)):
-            results['img'] = [to_tensor(img) for img in results['img']]
+            results['img'] = [self._to_tensor(img) for img in results['img']]
         else:
-            results['img'] = to_tensor(results['img'])
+            results['img'] = self._to_tensor(results['img'])
 
         return results
 
--- a/mmpose/datasets/pipelines/shared_transform.py
+++ b/mmpose/datasets/pipelines/shared_transform.py
@@ -33,7 +33,7 @@ class ToTensor:
 
     def __call__(self, results):
         to_tensor = (lambda x: torch.from_numpy(x.astype('float32')).permute(
-            2, 0, 1).to(self.device))
+            2, 0, 1).to(self.device) / 255.0)
         if isinstance(results['img'], (list, tuple)):
             results['img'] = [to_tensor(img) for img in results['img']]
         else:
--- a/mmpose/datasets/pipelines/shared_transform.py
+++ b/mmpose/datasets/pipelines/shared_transform.py
@@ -36,10 +36,10 @@ class ToTensor:
             2, 0, 1).to(self.device))
         if isinstance(results['img'], (list, tuple)):
             results['img'] = [
-                to_tensor(img.astype('float32')) for img in results['img']
+                to_tensor(img) for img in results['img']
             ]
         else:
-            results['img'] = to_tensor(results['img'].astype('float32'))
+            results['img'] = to_tensor(results['img'])
 
         return results
 
--- a/mmpose/apis/inference.py
+++ b/mmpose/apis/inference.py
@@ -49,7 +49,20 @@ def init_pose_model(config, checkpoint=None, device='cuda:0'):
     return model
 
 
+def _pipeline_gpu_speedup(pipeline, device):
+    """Load images to GPU and speed up the data transforms in pipelines.
 
+    Args:
+        pipeline: A instance of `Compose`.
+        device: A string or torch.device.
+
+    Examples:
+        _pipeline_gpu_speedup(test_pipeline, 'cuda:0')
+    """
+
+    for t in pipeline.transforms:
+        if isinstance(t, ToTensor):
+            t.device = device
 
 
 def _inference_single_pose_model(model,
@@ -195,7 +208,6 @@ def _inference_single_pose_model(model,
 
     batch_data = []
     for bbox in bboxes:
-
         # prepare data
         data = {
             'bbox':
@@ -218,6 +230,7 @@ def _inference_single_pose_model(model,
                 'flip_pairs': flip_pairs
             }
         }
+
         if use_multi_frames:
             # weight for different frames in multi-frame inference setting
             data['frame_weight'] = cfg.data.test.data_cfg.frame_weight_test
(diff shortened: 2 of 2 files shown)


mmpose/datasets/pipelines/loading.py
```diff
@@ -71,10 +71,10 @@ def __call__(self, results):
                 # The original results['img'] is assumed to be image(s) in BGR
                 # order, so we convert the color according to the arguments.
                 if results['img'].ndim == 3:
-                    results['img'] = mmcv.bgr2rgb(results['img'])
+                    results['img'] = results['img'][:, :, ::-1]
                 elif results['img'].ndim == 4:
                     results['img'] = np.concatenate(
-                        [mmcv.bgr2rgb(img) for img in results['img']], axis=0)
+                        [results['img'][:, :, ::-1] for img in results['img']], axis=0)
```
jin-s13, line 77: should be `[img[:, :, ::-1] for img in results['img']]` ?

chenxinfeng4: I'll fix it

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -27,11 +28,18 @@ class ToTensor:
...
+        to_tensor = (lambda x: torch.from_numpy(x.astype('float32')).permute(
+            2, 0, 1).to(self.device))
         if isinstance(results['img'], (list, tuple)):
-            results['img'] = [F.to_tensor(img) for img in results['img']]
+            results['img'] = [
+                to_tensor(img.astype('float32')) for img in results['img']
+            ]
         else:
-            results['img'] = F.to_tensor(results['img'])
+            results['img'] = to_tensor(results['img'].astype('float32'))
```
jin-s13, line 42: It seems that `.astype('float32')` is not necessary, because in `to_tensor`, x will first transformed to 'float32'.

mmpose/datasets/pipelines/loading.py
chenxinfeng4: I use `img[:, :, ::-1]` instead of `bgr2rgb(img)`, because the former is ligthweight for cpu.

mmpose/datasets/pipelines/shared_transform.py
chenxinfeng4: I'll fix it

jin-s13:
Thanks for contributing! @chenxinfeng4 
This PR has some lint issues. Please follow the instructions at https://github.com/open-mmlab/mmpose/blob/master/.github/CONTRIBUTING.md to setup the pre-commit.

chenxinfeng4:
I've tried the lint formating. Can you check it?

liqikai9 (approved):

chenxinfeng4:
image
This is test on `demo/top_down_video_demo_full_frame_without_det.py` with image_size=512x512. The FPS is a little bit slower (because the `img[:, :, ::-1]` for bgr2rgb is slower than `cv2.cvtColor`, but the former is cpu friendly.

The CPU workload was once the bottleneck in multiple GPUs inference in the origin case.

chenxinfeng4:
I'll improve the `demo/top_down_video_demo_full_frame_without_det.py` to 130fps with even less cpu workload （400%）.  CPU。这将是我下一个PR。
image

liqikai9:
Your improvement looks great! Do you have corresponding speed test results running on the GPU as well?

chenxinfeng4:
> 

Yeah. They were the tested in GPU NVIDIA3090. Beacuse both cases didn't show difference in GPU work load (arround 40-60% flucturating), so I didn't pose the GPU usage.

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -27,11 +28,16 @@ class ToTensor:
         results (dict): contain all information about training.
     """
 
+    def __init__(self, device='cpu'):
+        self.device = device
+
     def __call__(self, results):
+        to_tensor = (lambda x: torch.from_numpy(x.astype('float32')).permute(
```
liqikai9, line 35: It seems that when I used **multiple GPUs** and this [config](https://github.com/open-mmlab/mmpose/blob/master/configs/body/2d_kpt_sview_rgb_img/topdown_heatmap/coco/hrnet_w32_coco_256x192.py) for **testing**, the AP is near **zero**, which is not reasonable.

But when I used **single GPU**, the result is consistent . 

I also tried theses two cases:
1. Only keep your modification about in the `ToTensor`, the result with single GPU is correct while result in multiple GPUs is still near **zero**. 
2. Only keep your modification about in the `NormalizeTensor`, the results on single and multiple GPUa are both correct. 

I was wondering if it have something to do with the modifications you add inside `ToTensor`, which may cause some in place modification to the data implicitly.

chenxinfeng4: I didn't test in multiple GPU before. It seems unreasonable that the modification won't work in multiple GPU. What's the test file and command did you use? I will then check it and find the way out.

liqikai9: **UPDATE**: Previously I mistook the running path of MMPose. Now the results of single GPU and multiple GPUs are the same, but both got mAP near **zero**.

I use this config https://github.com/open-mmlab/mmpose/blob/master/configs/body/2d_kpt_sview_rgb_img/topdown_heatmap/coco/hrnet_w32_coco_256x192.py.  and download the corresponding checkpoint from this link https://download.openmmlab.com/mmpose/top_down/hrnet/hrnet_w32_coco_256x192-c78dce93_20200708.pth

1. For single GPU testing, I run:
```
PYTHONPATH="$(dirname $0)/..":$PYTHONPATH python tools/test.py configs/body/2d_kpt_sview_rgb_img/topdown_heatmap/coco/hrnet_w32_coco_256x192.py checkpoints/hrnet_w32_coco_256x192-c78dce93_20200708.pth
```

2. For multiple GPUs testing, I run:
```
GPUS=8 GPUS_PER_NODE=8 CPUS_PER_TASK=12 SRUN_ARGS='--quotatype=auto' ./tools/slurm_test.sh mm_human test_pr1320 configs/body/2d_kpt_sview_rgb_img/topdown_heatmap/coco/hrnet_w32_coco_256x192.py checkpoints/hrnet_w32_coco_256x192-c78dce93_20200708.pth
```

Another interesting phenomena is that when I run multi-GPU training using the same config and this command:
```
GPUS=8 GPUS_PER_NODE=8 CPUS_PER_TASK=12 SRUN_ARGS='--quotatype=auto' ./tools/slurm_train.sh mm_human train_pr1320 configs/body/2d_kpt_sview_rgb_img/topdown_heatmap/coco/hrnet_w32_coco_256x192.py work_dirs/
```
The evaluation results during training **looked just fine**., with mAP near 0.65 after 10 epochs. 

Considering the above testing results, this is a little bit strange.

liqikai9: When I change your modification in `ToTensor` back to `F.to_tensor`, the mAP is correct. 

So maybe it was the `to_tensor` function you add that somehow changes the data inplace.

chenxinfeng4:
Could that be the `inplace` in normalize?
```
results['img'] = F.normalize(
                results['img'], mean=self.mean, std=self.std, inplace=True)
```

I don't think the `to_tensor` has the issue. The `to_tensor` should copy the numpy to GPU without modification of the original numpy.
```
to_tensor = (lambda x: torch.from_numpy(x.astype('float32')).permute(
            2, 0, 1).to(self.device))
```

liqikai9:
I have tried these experiments.

1. I change 
```
results['img'] = F.normalize(
                results['img'], mean=self.mean, std=self.std, inplace=True)
```
to 

```
results['img'] = F.normalize(
                results['img'], mean=self.mean, std=self.std)
```
and keep your `to_tensor` function, the mAP is near zero.

2. I change your to_tensor function back to the previous way:
`results['img'] = F.to_tensor(results['img'])`
and keep your modification in normalize:
```
results['img'] = F.normalize(
                results['img'], mean=self.mean, std=self.std, inplace=True)
```
then the mAP is correct.

liqikai9:
Update: **I think I have found the problem.**

At first, the img array read after the pipeline `LoadImageFromFile` is in the range of **[0, 255]**.

It seems that in `F.to_tensor ` (check the doc [here](https://pytorch.org/vision/stable/generated/torchvision.transforms.ToTensor.html#torchvision.transforms.ToTensor))  the img array will be converted to the tensor in the range **[0.0, 1.0]** .
However, your `to_tensor` function just change the img array to tensor but didn't converted them to the range **[0.0, 1.0]**. 

I tried to change your to_tensor to:
```
to_tensor = (lambda x: torch.from_numpy(x.astype('float32')).permute(
            2, 0, 1).to(self.device)) / 255.0
```
without changing other places. The mAP is correct!!

Also, this can explain why during training, the mAP is near 0.7. I think during the training, the model can learn how to scale the img so the result is not zero.

liqikai9:
You may change your `to_tensor` function, and see check the speed test result on CPU.

chenxinfeng4:
Ohhh, it surprised me for the difference of `F.to_tensor` and `to_tensor`. I'll update that and check the speed again.

chenxinfeng4:
my to_tensor(),  65.35it/s
image

F.to_tensor(), 61.98it/s
image

mmpose/datasets/pipelines/shared_transform.py
```diff
@@ -27,11 +28,16 @@ class ToTensor:
...
+            2, 0, 1).to(self.device) / 255.0)
```
liqikai9, line 36: The `255.0` is a little bit hard-coded here. Is this always necessary? Because in the source code of `F.to_tensor`, 

https://github.com/pytorch/vision/blob/e9fb2a356384180d92dc60c49b4dd4b5b3c6ee8f/torchvision/transforms/functional.py#L152-L155

it seems that if the img is already located in [0.0, 1.0], it just returns the img tensor itself. 

BTW, `F.to_tensor` uses `div(255)`, I am not sure if this is faster. You may have a try!

chenxinfeng4: I've tested that in GPU and found the `division of 255.0` makes no difference in speed. The `division` is just similar to Normalization, which can also be speed up in GPU.

chenxinfeng4:
Why the merging is blocked? How can I fix it?

ly015:
> Why the merging is blocked? How can I fix it?

We are double-checking the optimized performance on our local devices. This feature will be merged soon and included in the next mmpose release.

mmpose/datasets/pipelines/loading.py
```diff
@@ -71,10 +71,10 @@ def __call__(self, results):
                 # The original results['img'] is assumed to be image(s) in BGR
                 # order, so we convert the color according to the arguments.
                 if results['img'].ndim == 3:
-                    results['img'] = mmcv.bgr2rgb(results['img'])
```
ly015, line 74: Numpy tensor indexing is faster than `mmcv.bgr2rgb` but the cost is 1e-6 sec/image level, which should be negligible. So I think we don't need to change the original code since it's more readable.

mmpose/datasets/pipelines/shared_transform.py
ly015: I suggest using the in-place operation `dev_(255)` to save memory.

liqikai9 (approved):

PR #1329, opened 2022-04-21
[Fix] Compatible with mme in model loading pretrained.

274869388:
## Motivation

The motivation of this PR is to compatible with mme in model loading pretrained. MME will call `init_weights` function without input parameter in the training script, which will overwrite the loaded pretrained model parameters. 

## Modification

- detector  
Modify the `pretrained` parameter to an attribute of the detector model. And `init_weights` function preferentially uses the incoming `pretrained`. If there are no input parameters, the attribute `self.pretrained` is used.

## BC-breaking (Optional)

No

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/models/detectors/associative_embedding.py
+++ b/mmpose/models/detectors/associative_embedding.py
@@ -65,7 +65,8 @@ class AssociativeEmbedding(BasePose):
         self.test_cfg = test_cfg
         self.use_udp = test_cfg.get('use_udp', False)
         self.parser = HeatmapParser(self.test_cfg)
-        self.init_weights(pretrained=pretrained)
+        self.pretrained = pretrained
+        self.init_weights()
 
     @property
     def with_keypoint(self):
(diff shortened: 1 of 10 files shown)


mmpose/models/detectors/associative_embedding.py
```diff
@@ -74,7 +75,9 @@ def with_keypoint(self):
 
     def init_weights(self, pretrained=None):
```
ly015, line 76: Do you mean that in MME `init_weights()` will be invoked separately after `__init__()`, which means `init_weights()` will be invoked twice during model building?

274869388: yes

ly015: Wouldn't that be weird since the model weights will be initialized twice?

274869388: This is only a temporary modification. In order to be compatible with mme, we have to initialize the parameters twice. It is recommended to inherit all model components from `BaseModule` and delete `init_weights()` in the future.

ly015 (approved):

jin-s13 (approved):

PR #1335, opened 2022-04-26
[Fix] Remove image file's existence check in `TopDownPoseTrack18VideoDataset`

liqikai9:
## Modification

1. Remove the image file's existence check in `TopDownPoseTrack18VideoDataset`, to avoid incorrect dataset content when the image files do not present locally, e.g, using Ceph as backend to load the image file.
2. Modify related unittests.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/datasets/datasets/top_down/topdown_posetrack18_video_dataset.py
+++ b/mmpose/datasets/datasets/top_down/topdown_posetrack18_video_dataset.py
@@ -206,15 +206,15 @@ class TopDownPoseTrack18VideoDataset(Kpt2dSviewRgbVidTopDownDataset):
                     continue
                 # the supporting frame index
                 support_idx = ref_idx + index
+                # clip the frame index to make sure that it does not exceed
+                # the boundings of frame indices
                 support_idx = np.clip(support_idx, 0, nframes - 1)
                 sup_image_file = cur_image_file.replace(
                     cur_image_name,
                     str(support_idx).zfill(self.ph_fill_len) + '.jpg')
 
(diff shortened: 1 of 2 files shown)


jin-s13 (approved):

mmpose/datasets/datasets/top_down/topdown_posetrack18_video_dataset.py
```diff
@@ -215,9 +214,6 @@ def _load_coco_keypoint_annotation_kernel(self, img_id):
                 if osp.exists(sup_image_file):
```
ly015, line 214: Please add some comments here to describe the relationship between `sup_image_file` and `cur_image_file` and when they will be used respectively?

liqikai9: The original purpose of these lines is to use `osp.exists` to check whether the `sup_image_file` exists locally. The frame indices have been checked above, which will make sure that they do not exceed the bounding of frame indices. But when the training data is on Ceph, every `sup_image_file` does not exist locally,  which will lead to a large amount of warnings. 

So, I  suggest that we can leave the file existence check in the pipeline like `LoadImageFromFile` instead of inside this dataset. And change these lines into:
```
image_files.append(sup_image_file)
```
Do you think this is proper?

ly015: Yes, we can check the image file's existence in the pipeline.

PR #1343, opened 2022-04-28
[Feature] add PVT backbone and configs

zengwang430521:
Add PVT backbone and configs.


ly015:
Please also update the README.md and README_CN.md.

jin-s13 (approved):

PR #1398, opened 2022-05-27
Adjust inference tracking min keypoints

pallgeuer:
## Motivation

The `get_track_id()` function in `inference_tracking.py` has a `min_keypoints` parameter that promises:
```
min_keypoints (int): Minimum number of keypoints recognized as person.
```
For the default value of 3, this means that all detected poses with 3 or more keypoints start a new track if it is not matched above a threshold to an existing track. The implementation has an off-by-one error though meaning in this case only detected poses with 4 or more keypoints will start a new track:
```
if np.count_nonzero(result['keypoints'][:, 1]) > min_keypoints:
```
This is unexpected/unclear for the user of the function.

## Modification

The obvious solution is to change `>` to `>=`, but this leads to a problem when `min_keypoints` is zero, so it needs to become:
```
if min_keypoints <= 0 or np.count_nonzero(result['keypoints'][:, 1]) >= min_keypoints:
```

## Use cases

One use case is that tracking by default is only performed by bbox IOU, so you would naturally expect the track ID of the 2D pose to be equivalent to the bounding box track ID, or want it to be (e.g. I do in my application). It is not equivalent however if pose detection returns too few keypoints. Passing `min_keypoints=0` makes it possible to achieve the described matching behaviour, which was not possible at all with the original code.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/apis/inference_tracking.py
+++ b/mmpose/apis/inference_tracking.py
@@ -182,7 +182,7 @@ def get_track_id(results,
             equivalent to an empty result list. Default: None
         next_id (int): The track id for the new person instance.
         min_keypoints (int): Minimum number of keypoints recognized as person.
-            default: 3.
+            0 means no minimum threshold required. Default: 3.
         use_oks (bool): Flag to using oks tracking. default: False.
         tracking_thr (float): The threshold for tracking.
         use_one_euro (bool): Option to use one-euro-filter. default: False.
@@ -218,7 +218,8 @@ def get_track_id(results,
         track_id, results_last, match_result = _track(result, results_last,
                                                       tracking_thr)
         if track_id == -1:
-            if np.count_nonzero(result['keypoints'][:, 1]) > min_keypoints:
+            if min_keypoints <= 0 or np.count_nonzero(
+                    result['keypoints'][:, 1]) >= min_keypoints:
                 result['track_id'] = next_id
                 next_id += 1
             else:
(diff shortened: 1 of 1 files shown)


mmpose/apis/inference_tracking.py
```diff
@@ -218,7 +218,8 @@ def get_track_id(results,
         track_id, results_last, match_result = _track(result, results_last,
                                                       tracking_thr)
         if track_id == -1:
-            if np.count_nonzero(result['keypoints'][:, 1]) > min_keypoints:
+            if min_keypoints <= 0 or np.count_nonzero(
```
ly015, line 221:
```suggestion
            if np.count_nonzero(result['keypoints'][:,1]) > min_keypoints:
```
I think we don't need to make ``min_keypoints==0`` a special case.

ly015:
@pallgeuer Thank you very much for your contribution!

mmpose/apis/inference_tracking.py
pallgeuer: I will try to reproduce it, but I am relatively sure that I received an occasional exception without the explicit handling of min_keypoints (months ago when I first discovered and patched this PR locally on my machine), but I can't 100% recall the situation. I think it may have been if a bounding box ended up with zero keypoint detections (rare case but possible) then result['keypoints'] ends up as None causing the following slice to error. Now that I think about it though, if that is truly the case then it would be better to explicitly handle the None case? I'm not sure, I'll have to see if I can make the exception happen again.

pallgeuer: I tried to reproduce the exception I described, but wasn't able to do so, so I guess the `min_keypoints == 0` special case is not required.

jin-s13 (approved):

PR #1402, opened 2022-05-30
Fix warning with torch.meshgrid

pallgeuer:
## Motivation

`torch.meshgrid()` has started raising a warning when not called with an explicit indexing parameter:
https://pytorch.org/docs/stable/generated/torch.meshgrid.html

## Modification
Provide `indexing='ij'` to all calls to `torch.meshgrid` that don't already specify indexing.

## Checklist

**Before PR**:

**After PR**:


mmpose/models/backbones/utils/transformer.py
```diff
@@ -796,7 +796,8 @@ def gen_encoder_output_proposals(self, memory, memory_padding_mask,
                 torch.linspace(
                     0, H - 1, H, dtype=torch.float32, device=memory.device),
                 torch.linspace(
-                    0, W - 1, W, dtype=torch.float32, device=memory.device))
+                    0, W - 1, W, dtype=torch.float32, device=memory.device),
+                indexing='ij')
```
jin-s13, line 800: This function is torch.linspace instead of torch.meshgrid. Is this change intended?

pallgeuer:
Refer to https://github.com/open-mmlab/mmdetection/pull/8090#issuecomment-1140937053 for discussion on PyTorch version support of `torch.meshgrid`.

mmpose/models/backbones/utils/transformer.py
pallgeuer: If you expand the context you can see it is a call to `torch.meshgrid` with two calls to `torch.linspace` as arguments.
```
            grid_y, grid_x = torch.meshgrid(
                torch.linspace(
                    0, H - 1, H, dtype=torch.float32, device=memory.device),
                torch.linspace(
-                    0, W - 1, W, dtype=torch.float32, device=memory.device))
+                    0, W - 1, W, dtype=torch.float32, device=memory.device),
                indexing='ij')
```

jin-s13: Aha, my fault.

mmpose/models/detectors/multiview_pose.py
```diff
@@ -53,7 +53,7 @@ def compute_grid(self, box_size, box_center, num_bins, device=None):
             grid_1D_x + box_center[0],
             grid_1D_y + box_center[1],
             grid_1D_z + box_center[2],
-        )
+            indexing='ij')
         grid_x = grid_x.contiguous().view(-1, 1)
```
ly015, line 57: The argument `indexing` has been introduced in PyTorch1.10.0 so we need to handle the compatibility of earlier versions. I think a version check could be added like:

```python
from packaging import version

if version.parse(torch.__verson__) < version.parse('1.10.0'):
    grid_x, grid_y, grid_z = torch.meshgrid(
            grid_1D_x + box_center[0],
            grid_1D_y + box_center[1],
            grid_1D_z + box_center[2])
else:
    grid_x, grid_y, grid_z = torch.meshgrid(
            grid_1D_x + box_center[0],
            grid_1D_y + box_center[1],
            grid_1D_z + box_center[2],
            indexing='ij')
```

ly015:
Thank you very much for your help!

mmpose/models/detectors/multiview_pose.py
pallgeuer: Yes, or seeing as there are multiple instances of `torch.meshgrid` one could write a wrapper for it that does the version check. Where would be an appropriate location for such a wrapper?

ly015: I think it's okay to directly use an if-else statement. If you prefer to use a wrapper, what about placing it at `mmpose/models/utils/misc.py`? Also, please note that `mmpose/models/backbones/utils/transformer.py` has been moved to 'mmpose/models/utils/transformer.py` this PR may need a rebase.

pallgeuer:
I have updated the pull request to now correctly deal with PyTorch versions above and below 1.10.0 using a wrapper. Can the PR be reviewed if all is well now?

mmpose/models/detectors/multiview_pose.py
pallgeuer: I have added a wrapper and adjusted

ly015 (approved):

PR #1436, opened 2022-06-20
Pavi logger hook

EvelynWang-0423:
## Motivation

PAVI startup optimization.

## Modification

PAVILoggerHook added in runtime.py


jin-s13:
Duplicated with #1434

PR #1438, opened 2022-06-21
[Enhancement] Test commands of mim in CI and clean pillow version check

liqikai9:
## Motivation

1. Test search command of mim in CI to ensure the format of metafile is correct. Refer to: https://github.com/open-mmlab/mmsegmentation/pull/1654 
2. Clean pillow version check. Refer to: https://github.com/open-mmlab/mmselfsup/pull/333

## Modification

1. Add `.github/workflows/test_mim.yml`.
3. Modify `.github/workflows/build.yml`

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/.github/workflows/test_mim.yml
+++ b/.github/workflows/test_mim.yml
@@ -39,6 +39,6 @@ jobs:
       - name: Install openmim
         run: pip install openmim
       - name: Build and install
-        run: rm -rf .eggs && pip install -e .
+        run: rm -rf .eggs && mim install -e .
       - name: test commands of mim
         run: mim search mmpose
--- a/.github/workflows/test_mim.yml
+++ b/.github/workflows/test_mim.yml
@@ -1,0 +1,44 @@
+name: test-mim
+
+on:
+  push:
+    paths:
+      - 'model-index.yml'
+      - 'configs/**'
+
+  pull_request:
+    paths:
+      - 'model-index.yml'
+      - 'configs/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build_cpu:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [3.7]
+        torch: [1.8.0]
+        include:
+          - torch: 1.8.0
+            torch_version: torch1.8
+            torchvision: 0.9.0
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Upgrade pip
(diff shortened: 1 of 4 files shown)


configs/body/2d_kpt_sview_rgb_img/topdown_heatmap/posetrack18/hrnet_w32_posetrack18_384x288.py
```diff
@@ -110,7 +110,7 @@
         type='NormalizeTensor',
         mean=[0.485, 0.456, 0.406],
         std=[0.229, 0.224, 0.225]),
-    dict(type='TopDownGenerateTarget', sigma=2),
+    dict(type='TopDownGenerateTarget', sigma=3),
```
ly015, line 113: Why are these 2 configs modified?

liqikai9: The input size 384x288 is usually with `sigmas = 3`. So here I change it. 

But I do not retrain these two configs and do not update the results on the `hrnet_posetrack18.md`.

.github/workflows/test_mim.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Upgrade pip
+        run: pip install pip --upgrade
+      - name: Install PyTorch
+        run: pip install torch==${{matrix.torch}}+cpu torchvision==${{matrix.torchvision}}+cpu -f https://download.pytorch.org/whl/torch_stable.html
+      - name: Install openmim
+        run: pip install openmim
+      - name: Build and install
+        run: rm -rf .eggs && pip install -e .
```
liqikai9, line 42: Here I used `pip install -e .` instead of `mim install -e .` in contrast to mmseg: https://github.com/open-mmlab/mmsegmentation/blob/master/.github/workflows/test_mim.yml#L47, because the latter seems to cause CI failure.

configs/body/2d_kpt_sview_rgb_img/topdown_heatmap/posetrack18/hrnet_w32_posetrack18_384x288.py
ly015: We'd better not change the config before it has been verified. Users may not be able to get the reported performance with new configs.

ly015: Also, it's irrelevant to this PR. We can update the configs in a separate PR.

liqikai9: Got it. Changed back to these two configs.

ly015 (approved):

PR #1481, opened 2022-07-11
Improve control of 3D visualisation of pose lifter

pallgeuer:
## Motivation

When visualising the 3D output poses coming from a pose lifter, it is currently not possible to control the visualisation height (pixels) and viewpoint angle, even though both are already implemented just under the hood, and would only need to be patched through. The former can be important for instance if a higher resolution output is desired, or as in my case, if you require it to have the exact same pixel height as the input image sequence (for post-processing reasons). I have also noticed  that the default viewpoint of the 3D visualisation is quite frontal, so in many cases the output looks correct, but if you look at it from a significantly different azimuth you can quickly tell it's not actually correct at all. In order to be able to properly evaluate and understand the outputs of various pose lifter models it is crucial to be able to visualise it from different viewing angles (I always use two views and plot them side by side for optimal understanding). Hence, why exposing `axis_azimuth` is quite important.

## Modification

I added arguments to allow `vis_height` and `axis_azimuth` to be set when calling `vis_3d_pose_result()`.
I also added some missing parameter documentation that I noticed was missing.

## BC-breaking

Nothing.

## Use cases

See motivation.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/apis/inference_3d.py
+++ b/mmpose/apis/inference_3d.py
@@ -364,7 +364,9 @@ def vis_3d_pose_result(model,
                        kpt_score_thr=0.3,
                        radius=8,
                        thickness=2,
+                       vis_height=400,
                        num_instances=-1,
+                       axis_azimuth=70,
                        show=False,
                        out_file=None):
     """Visualize the 3D pose estimation results.
@@ -459,7 +461,9 @@ def vis_3d_pose_result(model,
         thickness=thickness,
         pose_kpt_color=pose_kpt_color,
         pose_link_color=pose_link_color,
+        vis_height=vis_height,
         num_instances=num_instances,
+        axis_azimuth=axis_azimuth,
         show=show,
         out_file=out_file)
 
--- a/mmpose/models/detectors/pose_lifter.py
+++ b/mmpose/models/detectors/pose_lifter.py
@@ -296,6 +296,7 @@ class PoseLifter(BasePose):
                     thickness=2,
                     vis_height=400,
                     num_instances=-1,
+                    axis_azimuth=70,
                     win_name='',
                     show=False,
                     wait_time=0,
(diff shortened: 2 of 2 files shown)


jin-s13:
Thanks for your contribution！
It seems the unittest part fails. Could you please also take a look at it?

pallgeuer:
I had thought `PoseLifter` was the only model type that goes through that area of code, but `Interhand3D` indeed also does. So I have fixed that by adding the `axis_azimuth` parameter to `Interhand3D` as well. Note that the default value of `axis_azimuth` has to be `None` on the common pathway, because `PoseLifter` wants a default of 70 and `Interhand3D` wants a default of -115. I have preserved that behaviour.

ly015 (approved):

ly015:
Thanks for your contribution. I made minor modifications to the default value of the parameters.

pallgeuer:
Okay, so if I understand your commit correctly, the default internal value of `axis_azimuth` is now always 70, but the InterHand3D demo (the only place the vis method is ever called for InterHand?) specifies an explicit value of -115? And the InterHand3D class also tries to enforce a default value of -115, but this value is generally ignored anyway because the general vis method above it in the callstack already applied a default of 70 if no explicit `axis_azimuth` is specified?

ly015:
Yes, 

> Okay, so if I understand your commit correctly, the default internal value of `axis_azimuth` is now always 70, but the InterHand3D demo (the only place the vis method is ever called for InterHand?) specifies an explicit value of -115? And the InterHand3D class also tries to enforce a default value of -115, but this value is generally ignored anyway because the general vis method above it in the callstack already applied a default of 70 if no explicit `axis_azimuth` is specified?

Yes, that's correct. Although in our demos we always invoke the top-level visualization interfaces like `vis_3d_pose_result()`, it is not prohibited to directly use the lower-level functions like `model.show_results()`. So I think it also makes sense to keep a default parameter value for the latter. And we are also working on simplifying the encapsulation of inference and visualization APIs, which is a little bit deep indeed.

BTW, sorry for the hasty merge of this PR. If you have any concerns, we can open a new PR to further polish it.

pallgeuer:
No, I'm happy with the current merge, now that I have understood it.

PR #1500, opened 2022-07-21
[Refactor] Reorganize HeatmapHead and BaseHead class in order to develop RegressionHead

Tau-J:
## Motivation

Refactor `HeatmapHead` and `BaseHead` class

## Modification

Reorganized `decode`, `_get_in_channels`, `_transform_inputs` into `BaseHead`.
Modified `decode` for supporting `heatmap-based`, `regression-based` and `integral-regression` methods. Support heatmap visualization config via `test_cfg`.
Saved `_to_numpy` as a util function in /core/utils/tensor_utils.py

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/models/heads/base_head.py
+++ b/mmpose/models/heads/base_head.py
@@ -4,7 +4,7 @@ from typing import Tuple, Union
 
 import torch
 import torch.nn.functional as F
-from mmengine.data import InstanceData, PixelData
+from mmengine.data import InstanceData
 from mmengine.model import BaseModule
 from torch import Tensor
 
@@ -132,9 +132,5 @@ class BaseHead(BaseModule, metaclass=ABCMeta):
                 data_sample.pred_instances = InstanceData()
             data_sample.pred_instances.keypoints = keypoints
             data_sample.pred_instances.keypoint_scores = scores
-
-            # Store the heatmap predictions in the data sample
-            if 'pred_fileds' not in data_sample:
-                data_sample.pred_fields = PixelData()
 
         return batch_data_samples
--- a/mmpose/models/heads/base_head.py
+++ b/mmpose/models/heads/base_head.py
@@ -99,31 +99,17 @@ class BaseHead(BaseModule, metaclass=ABCMeta):
 
         if isinstance(batch_outputs, Tensor):
             batch_outputs_np = _to_numpy(batch_outputs)
-            batch_heatmaps_np = batch_outputs_np
-
-            head_type = type(self).__name__.lower()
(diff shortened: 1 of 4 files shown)


mmpose/models/heads/base_head.py
```diff
@@ -29,3 +33,126 @@ def predict(self, feats: Tuple[Tensor], batch_data_samples: OptSampleList,
...
+        if batch_data_samples is None:
+            raise ValueError(
+                '`batch_data_samples` is required to decode keypoitns.')
+
+        if isinstance(batch_outputs, Tensor):
+            batch_outputs_np = _to_numpy(batch_outputs)
+            batch_heatmaps_np = batch_outputs_np
+
+            head_type = type(self).__name__.lower()
+            if 'regression' in head_type:
```
ly015, line 105: If we move 'decode()' to the base class, I think it's better to assign `pred_fields.heatmap` in `predict()` instead of `decode()` since it's more head-specific.

ly015: Besides, it may be risky to control the behavior by the class name. This won't be a problem if we assign `pred_fields.heatmap` in subclasses.

Tau-J: Good suggestion.

mmpose/models/heads/heatmap_heads/heatmap_head.py
```diff
@@ -304,56 +242,18 @@ def forward(self, feats: Tuple[Tensor]) -> Tensor:
...
-            # Store the heatmap predictions in the data sample
-            if 'pred_fileds' not in data_sample:
-                data_sample.pred_fields = PixelData()
-            data_sample.pred_fields.heatmaps = heatmaps
+        for heatmaps, data_sample in zip(batch_heatmaps, batch_data_samples):
+            if vis_heatmaps:
+                data_sample.pred_fields.heatmaps = heatmaps
 
-        return batch_data_samples
+        return pred_coords
```
ly015: `pred_coords` can be confusing because it's actually a list of data samples (the input data samples with `pred_instances` added in place) instead of just coordinates. And maybe 'output_heatmaps' is better than 'vis_heatmaps'.

```python
if test_cfg.get('output_heatmaps', True):
    for heatmaps, data_sample in zip(batch_heatmaps, batch_data_samples):
        if 'pred_fields' not in data_sample:
            data_sample.pred_fields = PixelData()
        data_sample.pred_fields.heatmaps = heatmaps
```

ly015 (approved):

Issue #1502, opened 2022-07-22
Cannot play track demo with non-COCO format keypoints in oks_iou mode


daixinghome:
**Describe the bug**

In the latest commit, I find that the track demo script can only make visualization of keypoints defined by COCO. If I use the other definition with different keypoints number (not equal to 17), it will always fail.

**Reproduction**

For example, if you run the following code in default environment, you cannot be success.

```
python3 demo/bottom_up_pose_tracking_demo.py \
    configs/body/2d_kpt_sview_rgb_img/associative_embedding/aic/higherhrnet_w32_aic_512x512.py \
    https://download.openmmlab.com/mmpose/bottom_up/higher_hrnet32_aic_512x512-9a674c33_20210130.pth \
    --kpt-thr 0.6 \
    --smooth \
    --use-oks-tracking \
    --video-path demo/resources/demo.mp4 \
    --out-video-root vis_results
```

**Error traceback**

```
  File "demo/bottom_up_pose_tracking_demo.py", line 208, in <module>
    main()
  File "demo/bottom_up_pose_tracking_demo.py", line 166, in main
    pose_results, next_id = get_track_id(
  File "/home/xing.dai/workspace/mmpose/mmpose/apis/inference_tracking.py", line 226, in get_track_id
    track_id, results_last, match_result = _track(
  File "/home/xing.dai/workspace/mmpose/mmpose/apis/inference_tracking.py", line 106, in _track_by_oks
    oks_score = oks_iou(pose, pose_last, area, area_last, sigmas=sigmas)
  File "/home/xing.dai/workspace/mmpose/mmpose/core/post_processing/nms.py", line 81, in oks_iou
    e = (dx**2 + dy**2) / vars / ((a_g + a_d[n_d]) / 2 + np.spacing(1)) / 2
ValueError: operands could not be broadcast together with shapes (14,) (17,)
```

**Bug fix**

The key error is that in the code of oks_iou, no `sigmas` is specified in the tracking code:
https://github.com/open-mmlab/mmpose/blob/e37e758ab743ea71b6c4c4b7f3ebf0d6a243e1c0/mmpose/apis/inference_tracking.py#L218

So it always use default sigmas of COCO dataset.
https://github.com/open-mmlab/mmpose/blob/e37e758ab743ea71b6c4c4b7f3ebf0d6a243e1c0/mmpose/core/post_processing/nms.py#L51

If we get the sigmas from DatasetInfo and pass the variable to oks_iou function, it will work.
[...]

PR #1504, opened 2022-07-22
[Fix] fix bug of non-COCO dataset track demo

daixinghome:
## Motivation

Fix: https://github.com/open-mmlab/mmpose/issues/1502

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/apis/inference_tracking.py
+++ b/mmpose/apis/inference_tracking.py
@@ -36,7 +36,7 @@ def _compute_iou(bboxA, bboxB):
     return iou
 
 
-def _track_by_iou(res, results_last, thr):
+def _track_by_iou(res, results_last, thr, **kwargs):
     """Get track id using IoU tracking greedily.
 
     Args:
@@ -75,7 +75,7 @@ def _track_by_iou(res, results_last, thr):
     return track_id, results_last, match_result
 
 
-def _track_by_oks(res, results_last, thr):
+def _track_by_oks(res, results_last, thr, sigmas, **kwargs):
     """Get track id using OKS tracking greedily.
 
     Args:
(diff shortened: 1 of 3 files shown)


Ben-Louis:
@daixinghome Thank you very much for your contribution! Could you please sign the CLA?

jin-s13 (approved):

daixinghome:
>

daixinghome:
Sorry for the mistake. Now it's ok~

mmpose/apis/inference_tracking.py
```diff
@@ -36,7 +36,7 @@ def _compute_iou(bboxA, bboxB):
     return iou
 
 
-def _track_by_iou(res, results_last, thr):
+def _track_by_iou(res, results_last, thr, **kwargs):
```
ly015, line 39: I would recommend not using ``**kwargs`` but it seems there is no other simple solution here. @jin-s13 Do you have any suggestions?

Ben-Louis: Maybe we can use `functools.partial` to unify the interfaces

ly015: > Maybe we can use `functools.partial` to unify the interfaces

Could you please give an example?

Ben-Louis: > Could you please give an example?

```python
def _track_by_iou(res, results_last, thr): 
    # track by iou

def _track_by_oks(res, results_last, thr, sigmas): 
    # track by oks

_track = partial(_track_by_oks, sigmas=dataset_sigmas) if use_oks else _track_by_iou
```

if `use_oks` is True, then `_track` acts like the following function
```python
def func(res, results_last, thr): 
    return _track_by_oks(res, results_last, thr, dataset_sigmas)
```

daixinghome: Good idea. I've already modified with this method :).

PR #1514, opened 2022-07-26
[Refactor] add Regression, IPR, SimCC method

Tau-J:
## Motivation

Add `RegressionHead` and `IntegralRegressionHead`.
Add `SimCCHead`.
Add `RegressionLabel` codec for regression-based methods.
Add `SimCCLabel` codec for simcc-based methods.
Add `DSNTLoss`, `KLDiscretLoss`, `KLLossWithLabelSmoothing`.
Add unittest for all modules above except loss.

## Checklist

**Before PR**:

**After PR**:


mmpose/core/keypoint/codecs/regression_label.py
```diff
@@ -0,0 +1,77 @@
...
+        super().__init__()
+
+        self.input_size = input_size
+
+    def encode(
+        self,
+        keypoints: np.ndarray,
+        keypoints_visible: Optional[np.ndarray] = None
+    ) -> Tuple[np.ndarray, np.ndarray]:
+        """Encoding keypoints from input image space to normalized space."""
```
ly015, line 34: Please also describe args and returns

mmpose/datasets/pipelines/topdown_transforms.py
```diff
@@ -219,7 +219,8 @@ def __repr__(self) -> str:
 
 @TRANSFORMS.register_module()
 class TopdownGenerateRegressionLabel(BaseTransform):
```
ly015, line 221: Do you think it would be better to combine `TopdownGenerateHeatmap` and `TopdownGenerateRegressionLabel` into a  unified transform (e.g. `TopdownGenerateTarget`) or not?

mmpose/datasets/pipelines/topdown_transforms.py
```diff
@@ -284,3 +284,63 @@ def __repr__(self) -> str:
         repr_str += ('(use_dataset_keypoint_weights='
                      f'{self.use_dataset_keypoint_weights})')
         return repr_str
+
+
+@TRANSFORMS.register_module()
+class TopdownGenerateSimCCLabel(BaseTransform):
```
ly015, line 290: I suggest using `TopdownGenerateRegressionLabel` for SimCC. `TopdownGenerateRegressionLabel` should be a general data transform that can support different regression-based methods by giving different encoders.

mmpose/datasets/pipelines/topdown_transforms.py
Tau-J: It seems to be a good idea :) I think it's better to fold them as two util functions and use `TopdownGenerateTarget` to call

mmpose/models/heads/heatmap_heads/heatmap_head.py
```diff
@@ -306,33 +306,10 @@ def _load_state_dict_pre_hook(self, state_dict, prefix, local_meta, *args,
...
-                    if idx < len(self.conv_layers):
-                        # final_layer.n.xxx -> conv_layers.n.xxx
-                        k_new = 'conv_layers.' + '.'.join(k_parts[1:])
-                    else:
-                        # final_layer.n.xxx -> final_layer.xxx
-                        k_new = 'final_layer.' + k_parts[2]
-                else:
-                    # final_layer.xxx remains final_layer.xxx
-                    k_new = k
-            elif k_parts[0] == 'loss':
```
ly015, line 335: These codes are necessary for compatibility with old version checkpoints.

mmpose/datasets/pipelines/topdown_transforms.py
Tau-J: In fact, I'm not sure if it's appropriate to regard SimCC as a heatmap-based method as well. It locates keypoints by doing classification on 1d vectors, both the outputs and targets are 1d vectors.

ly015: Maybe we can use one transform `TopdownGenerateTarget` for all methods including heatmap-based, regression-based, and SimCC (whatever it belongs to)? The target name (e.g. 'heatmaps', 'reg_labels') can be controlled by a parameter.

And maybe we can use a term like `keypoint_labels` for both the regression methods and SimCC to avoid confusion?

mmpose/datasets/pipelines/topdown_transforms.py
```diff
@@ -142,8 +142,8 @@ def __repr__(self) -> str:
 
 
 @TRANSFORMS.register_module()
-class TopdownGenerateHeatmap(BaseTransform):
-    """Encode keypoints into heatmaps.
+class TopdownGenerateTarget(BaseTransform):
```
ly015, line 145: I think it's better not to directly entangle `TopdownGenerateTarget` and the codec type. We can add a parameter `target_type` to `TopdownGenerateTarget` to control how to handle the encoded items. I think the logic will be more clear and it would be easier to add new codecs.

And we can move the multi-level heatmap part into a codec wrapper (a special codec, I will implement it later in another  PR) to further unify the processing of different target types and simplify `TopdownGenerateTarget`.

So basically it will be something like:

```python
class TopdownGenerateTarget(BaseTransform):
    """Encode keypoints into the form of the model learning target like heatmaps
    or keypoint labels.

    Required Keys:

        - keypoints
        - keypoints_visible
        - dataset_keypoint_weights

    Added Keys (depends on the args):
        - heatmaps
        - keypoint_labels
        - keypoint_x_labels
        - keypoint_y_labels
        - keypoint_weights

    Args:
        encoder (dict | list[dict]): The codec config for keypoint encoding
        target_type (str): The type of the encoded form of the keypoints.
            Should be one of the following options:

            - ``'heatmap'``: The encoded should be instance-irrelevant
                heatmaps and will be stored in ``results['heatmaps']``
            - ``'keypoint_label'``: The encoded should be instance-level
                labels and will be stored in ``results['keypoint_label']``
            - ``'keypoint_xy_label'``: The encoded should be instance-level
                labels in x-axis and y-axis respectively. They will be stored
                in ``results['keypoint_x_label']`` and
                ``results['keypoint_y_label']``
    """

    def __init__(self,
                 encoder: MultiConfig,
                 target_type: str,
                 use_dataset_keypoint_weights: bool = False) -> None:
        super().__init__()
        self.target_type = target_type
        self.encoder_cfg = encoder
        self.use_dataset_keypoint_weights = use_dataset_keypoint_weights

        self.encode = KEYPOINT_CODECS.build(self.encoder_cfg)

    def transform(self,
                  results: Dict) -> Optional[Union[Dict, Tuple[List, List]]]:

        if self.target_type == 'heatmap':
            heatmaps, keypoint_weights = self.encoder.encode(
                keypoints=results['keypoints'],
                keypoints_visible=results['keypoints_visible'])

            results['heatmaps'] = heatmaps
            results['keypoint_weights'] = keypoint_weights

        elif self.target_type == 'keypoint_label':
            keypoint_label, keypoint_weights = self.encoder.encode(
                keypoints=results['keypoints'],
                keypoints_visible=results['keypoints_visible'])

            results['keypoint_label'] = keypoint_label
            results['keypoint_weights'] = keypoint_weights

        elif self.target_type == 'keypoint_xy_label':
            x_label, y_label, keypoint_weights = self.encoder.encode(
                keypoints=results['keypoints'],
                keypoints_visible=results['keypoints_visible'])

            results['keypoint_x_label'] = x_label
            results['keypoint_y_label'] = y_label
            results['keypoint_weights'] = keypoint_weights
        
        else:
            raise ValueError(f'Invalid target type {self.target_type}')

        # multiply meta keypoint weight
        if self.use_dataset_keypoint_weights:
            results['keypoint_weights'] *= results['dataset_keypoint_weights']

        return results

    def __repr__(self) -> str:
        """print the basic information of the transform.

        Returns:
            str: Formatted string.
        """
        repr_str = self.__class__.__name__
        repr_str += (f'(encoder={str(self.encoder_cfg)}, ')
        repr_str += (f'(target_type={str(self.target_type)}, ')
        repr_str += ('use_dataset_keypoint_weights='
                     f'{self.use_dataset_keypoint_weights})')
        return repr_str
```

What do you think?

ly015 (approved):

PR #1544, opened 2022-08-05
[Refactor] support flip test for heatmap heads

ly015:
## Motivation

Support test time augmentation by flipping for heatmap heads (`HeatmapHead`, `CPMHead`, `MSPNHead` and `ViPNASHead`)

## Checklist

**Before PR**:

**After PR**:


mmpose/models/utils/tta.py
```diff
@@ -0,0 +1,48 @@
...
+        B, C, H, W = heatmaps.shape
+        assert len(flip_indices) * 3 == C
+        heatmaps = heatmaps.view(B, C // 3, 3, H, W)
+        heatmaps = heatmaps[:, flip_indices].flip(-1)
+        heatmaps[:, :, 1] = -heatmaps[:, :, 1]
+        heatmaps = heatmaps.view(B, C, H, W)
+    else:
+        raise ValueError(f'Invalid flip_mode value "{flip_mode}"')
+
+    if shift_heatmap:
```
Tau-J, line 45: Could provide any corresponding reference in the previous codes? I'm not sure whether it is appropriate to shift heatmaps after udp_combined results

ly015: Yes, `shift_heatmap` is not used for any UDP model.

Tau-J: How about setting a limit on it in case of misusing?

ly015: Doesn't seem easy to do so. The function `flip_heatmap` and even the head are not supposed to be able to directly check the codec type. It won't be a problem, I think. In most cases, users will follow our configs on these settings.

Tau-J (approved):

liqikai9:
I have modified the example config `configs/tests/hrnet_w32_coco_256x192.py`. And now we can run the command like:
`python tools/test.py configs/tests/hrnet_w32_coco_256x192.py ~/checkpoints/hrnet_w32_coco_256x192-c78dce93_20200708.pth`
or 
`./tools/slurm_test.sh mm_dev test configs/tests/hrnet_w32_coco_256x192.py ~/checkpoints/hrnet_w32_coco_256x192-c78dce93_20200708.pth`
to test the config with one GPU or multiple GPUs. 

I modify these scripts' mode: `tools/*.sh` to make them executable, changing their modes from 100644 to 100755. 

The mAP results are aligned with these items listed here: `https://github.com/open-mmlab/mmpose/blob/master/configs/body/2d_kpt_sview_rgb_img/topdown_heatmap/coco/hrnet_coco.md`, with about 1e-4 difference with the results in `hrnet_coco.md`.

liqikai9 (approved):

ly015:
Changing file modes seems unnecessary. You can always execute them by `sh xxx` or `python xxx`.

PR #1562, opened 2022-08-11
[Refactor] Verify RLE on MPII

Tau-J:
## Motivation
support pckh on mpii dataset

## Modification

modified `base_coco_style_dataset.py` to add headbox_file
fix bug on mpii dataset

## Checklist

**Before PR**:

**After PR**:


mmpose/datasets/datasets/body/mpii_dataset.py
```diff
@@ -115,16 +131,25 @@ def _load_annotations(self) -> List[dict]:
                 keypoints = np.zeros((1, num_keypoints, 2), dtype=np.float32)
```
Tau-J, line 131: Now I'm curious about why we use dummy data as keypoints on test_mode, which result in a incorrect PCKh computation.

liqikai9: I suppose this is to be aligned with the previous version: https://github.com/open-mmlab/mmpose/blob/master/mmpose/datasets/datasets/top_down/topdown_mpii_dataset.py#L113.

Intuitively, the ground truth keypoints positions are unknown and also unnecessary when `test_mode` is `True`.

Tau-J: But in previoius version, the metric `PCKh` is calculated in `evaluation()` in mpii_dataset, which loads keypoints and head box from the `.mat` file individually. But right now, we need to load the head boxes infomation and pass them in gt_instances. Therefore, maybe we should change the codes here.

mmpose/datasets/datasets/base/base_coco_style_dataset.py
```diff
@@ -63,6 +63,7 @@ class BaseCocoStyleDataset(BaseDataset):
     def __init__(self,
                  ann_file: str = '',
                  bbox_file: Optional[str] = None,
+                 headbox_file: Optional[str] = None,
```
ly015, line 66: If so far `headbox_file` is only for the mpii dataset, I suggest only adding it in ``MpiiDataset``

ly015:
Maybe use `ori_keypoints`? MMCV uses `ori_shape` so we can use the same convention.
https://github.com/open-mmlab/mmcv/blob/dev-2.x/mmcv/transforms/loading.py#L81

configs/tests/res50_mpii_256x256.py
```diff
@@ -0,0 +1,94 @@
+_base_ = [
```
ly015, line 1: Please move the verified configs to proper locations like `configs/body2d_keypoints/topdown_regression/mpii/'.

ly015 (approved):

PR #1579, opened 2022-08-18
[Fix] Keep gt keypoints unchanged in val/test pipelines

ly015:
## Motivation

Keep 'keypoints' unchanged in val_pipeline and test_pipelin. Otherwise, it will lead to incorrect evaluation results.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/datasets/datasets/body/mpii_trb_dataset.py
+++ b/mmpose/datasets/datasets/body/mpii_trb_dataset.py
@@ -7,6 +7,7 @@ import numpy as np
 from mmengine.utils import check_file_exist
 
 from mmpose.registry import DATASETS
+from mmpose.structures.bbox import bbox_cs2xyxy
 from ..base import BaseCocoStyleDataset
 
 
@@ -122,6 +123,7 @@ class MpiiTrbDataset(BaseCocoStyleDataset):
             center = np.array(ann['center'], dtype=np.float32)
             scale = np.array([ann['scale'], ann['scale']],
                              dtype=np.float32) * pixel_std
+            bbox = bbox_cs2xyxy(center, scale)
 
             # keypoints in shape [1, K, 2] and keypoints_visible in [1, K]
             _keypoints = np.array(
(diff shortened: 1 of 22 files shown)


mmpose/evaluation/metrics/keypoint_2d_metrics.py
```diff
@@ -77,9 +77,9 @@ def process(self, data_batch: Sequence[dict],
             # ground truth data_info
             gt = data['data_sample']['gt_instances']
             # ground truth keypoints coordinates, [1, K, D]
-            gt_coords = gt['ori_keypoints']
+            gt_coords = gt['keypoints']
             # ground truth keypoints_visible, [1, K, 1]
-            mask = gt['ori_keypoints_visible'].astype(bool).reshape(1, -1)
+            mask = gt['keypoints_visible'].astype(bool).reshape(1, -1)
```
Ben-Louis, line 82: I think it is more reasonable to convert `mask` to boolean array here instead of in `_calc_distances`. This matches intuition and descriptions better. This conversion should also be applied to `mask` in the other Metrics.

mmpose/datasets/datasets/body/mpii_trb_dataset.py
```diff
@@ -129,32 +129,20 @@ def _load_annotations(self) -> List[dict]:
...
-                'ori_keypoints':
-                keypoints.copy(),
-                'ori_keypoints_visible':
-                keypoints_visible.copy(),
-                'iscrowd':
-                ann['iscrowd'],
+                'id': ann['id'],
+                'img_id': img_id,
+                'img_path': img_path,
+                'bbox_center': center,
```
Ben-Louis, line 139: `bbox` is necessary for `data_info`

mmpose/datasets/datasets/body/mpii_trb_dataset.py
```diff
@@ -122,39 +123,29 @@ def _load_annotations(self) -> List[dict]:
             center = np.array(ann['center'], dtype=np.float32)
             scale = np.array([ann['scale'], ann['scale']],
                              dtype=np.float32) * pixel_std
```
Ben-Louis, line 125:
```suggestion
            # center, scale in shape [1, 2] and bbox in [1, 4]
            center = np.array([ann['center']], dtype=np.float32)
            scale = np.array([[ann['scale'], ann['scale']]],
                             dtype=np.float32) * pixel_std
```
`center` and `scale` should have shape [1, 2]

Tau-J (approved):

PR #1603, opened 2022-08-28
[Refactor] refactor Associative Embedding codec

ly015:
## Motivation

Add codecs for Associative Embedding.

## Checklist

**Before PR**:

**After PR**:


mmpose/codecs/utils/refinement.py
```diff
@@ -0,0 +1,165 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from itertools import product
+
+import numpy as np
+
+from .post_processing import gaussian_blur
+
+
+def refine_keypoints(keypoints: np.ndarray,
```
Tau-J, line 9: Just as a reminder, processing methods like these should be given in the documentation afterwards to facilitate user learning

ly015: These functions will be automatically included in the `apis` section of the documentation, with their signatures and docstrings showed on the page. Do you think we need extra documents for them?

PR #1637, opened 2022-09-06
[Doc] Refine inference.md, prepare_dataset.md, and train_and_test.md

Ben-Louis:
## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/docs/en/user_guides/inference.md
+++ b/docs/en/user_guides/inference.md
@@ -1,20 +1,20 @@
 # Inference with existing models
 
 MMPose provides hundreds of pre-trained models for pose estimation in [Model Zoo](https://mmpose.readthedocs.io/en/1.x/modelzoo.html).
-This note will show **how to inference**, which means use trained models to run pose estimation on given images or videos.
+This note will show **how to perform inference**, which means running pose estimation on given images or videos with trained models.
 
-As for how to test existing models on standard datasets, please see this [guide](./train_and_test.md#test)
+As for how to test existing models on standard datasets, please see this [guide](./train_and_test.md#test).
 
-In MMPose, a model is defined by a configuration file and existing model parameters are save in a checkpoint file.
+In MMPose, a model is defined by a configuration file and existing model parameters are saved in a checkpoint file.
 
-To start with, we recommend HRNet model with [this configuration file](configs/body_2d_keypoint/topdown_heatmap/coco/td-hm_hrnet-w32_8xb64-210e_coco-256x192.py) and [this checkpoint file](https://download.openmmlab.com/mmpose/top_down/hrnet/hrnet_w32_coco_256x192-c78dce93_20200708.pth). It is recommended to download the checkpoint file to `checkpoints` directory.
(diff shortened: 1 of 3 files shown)


docs/en/user_guides/inference.md
```diff
@@ -1,20 +1,20 @@
 # Inference with existing models
```
liqikai9, line 1: How about changing to `Inference with trained models`?

docs/en/user_guides/prepare_datasets.md
```diff
@@ -80,7 +93,7 @@ An example of the dataset config is as follows.
 4. `type`: 'upper' or 'lower', will be used in data augmentation.
 5. `swap`: indicates the 'swap pair' (also known as 'flip pair'). When applying image horizontal flip, the left part will become the right part. We need to flip the keypoints accordingly.
 
-`skeleton_info` contains the information about the keypoint connectivity, which is used for visualization.
+`skeleton_info` contains information about the keypoint connectivity, which is used for visualization.
 
 `joint_weights` assigns different loss weights to different keypoints.
```
Tau-J, line 99:
```Python
keypoint_info={
        0:
        dict(name='nose', id=0, color=[51, 153, 255], type='upper', swap=''),
        1:
        dict(
            name='left_eye',
            id=1,
            color=[51, 153, 255],
            type='upper',
            swap='right_eye'),
        2:
        dict(
            name='right_eye',
            id=2,
            color=[51, 153, 255],
            type='upper',
            swap='left_eye'),
        3:
        dict(
            name='left_ear',
            id=3,
            color=[51, 153, 255],
            type='upper',
            swap='right_ear'),
```

I think this list is too long to read and it't not necessary to show the whole `keypoint_info` list in this example

liqikai9: Agreed. I think we can omit the middle part of `keypoint_info` as well as `skeleton_info`.

docs/en/user_guides/train_and_test.md
```diff
@@ -1,5 +1,22 @@
 # Training and Test
```
Tau-J, line 1: Why do you use `Training and Test`, this is so weird..

docs/en/user_guides/inference.md
Ben-Louis: I think the range of existing models is broader, which includes trained models, pruned models, manually converted models, etc. All of them may run inference with MMPose.

docs/en/user_guides/prepare_datasets.md
Ben-Louis: Good idea

docs/en/user_guides/train_and_test.md
Ben-Louis: I think it is because this doc uses the corresponding [MMClassification doc](https://github.com/open-mmlab/mmclassification/blob/1.x/docs/en/user_guides/train_test.md) as a reference. 

Do you suggest changing 'Test' to 'Testing'?

Tau-J: Personally I prefer `Train & Test` like MMDetection

:D

image

liqikai9 (commented):
Better to add the corresponding link to `mmengine` tutorial in the end of `prepare_datasets.md` and remove the `TODO`

liqikai9 (approved):

PR #1647, opened 2022-09-08
[Refactor] Modify some scripts to fit `MultilevelPixelData`

liqikai9:
## Motivation

1. Unify the `multiscale`, `multi-scale`, `multi-level` to `multilevel`; add `multilevel_gt_fields` in `PoseDataSample`;
2. Modify `MSPNHead` to fit `MultilevelPixelData`;
3. Fix some bugs in `MultilevelPixelData` add unittests for `MultilevelPixelData`;
4. Modify the MSPNd and RSN configs according to the above modificatios.

All unitetsts passed locally.
The test accuracy for `configs/body_2d_keypoint/topdown_heatmap/coco/td-hm_2xmspn50_8xb32-210e_coco-256x192.py` is verified, will also verify other modified configs.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/configs/body_2d_keypoint/topdown_heatmap/coco/td-hm_2xrsn50_8xb32-210e_coco-256x192.py
+++ b/configs/body_2d_keypoint/topdown_heatmap/coco/td-hm_2xrsn50_8xb32-210e_coco-256x192.py
@@ -30,11 +30,15 @@ auto_scale_lr = dict(base_batch_size=256)
 default_hooks = dict(checkpoint=dict(save_best='coco/AP', rule='greater'))
 
 # codec settings
-codec = dict(
-    type='MegviiHeatmap',
-    input_size=(192, 256),
-    heatmap_size=(48, 64),
-    kernel_size=5)
+# multiple kernel_sizes of heatmap gaussian for 'Megvii' approach.
+kernel_sizes = [15, 11, 9, 7] + [11, 9, 7, 5]
+codec = [
+    dict(
+        type='MegviiHeatmap',
+        input_size=(192, 256),
+        heatmap_size=(48, 64),
+        kernel_size=kernel_size) for kernel_size in kernel_sizes
+]
 
 # model settings
 model = dict(
@@ -71,7 +75,7 @@ model = dict(
                 use_target_weight=True,
                 loss_weight=1.)
         ]) * 2,
-        decoder=codec),
+        decoder=codec[-1]),
     test_cfg=dict(
         flip_test=True,
         flip_mode='heatmap',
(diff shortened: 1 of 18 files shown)


mmpose/datasets/transforms/formatting.py
```diff
@@ -146,6 +148,13 @@ def transform(self, results: dict) -> dict:
                 gt_fields.set_field(results[key], packed_key)
         data_sample.gt_fields = gt_fields.to_tensor()
 
+        # pack multilevel fields
```
ly015, line 151: I think we better use `gt_fields` and `pred_fields` for both single-level and mulit-level cases, and choose its data type between `PixelData` and `MultilevelPixelData` according to the data (np.ndarray or list) to pack.

configs/body_2d_keypoint/topdown_heatmap/coco/td-hm_2xrsn50_8xb32-210e_coco-256x192.py
```diff
@@ -30,11 +30,15 @@
 default_hooks = dict(checkpoint=dict(save_best='coco/AP', rule='greater'))
 
 # codec settings
-codec = dict(
-    type='MegviiHeatmap',
-    input_size=(192, 256),
-    heatmap_size=(48, 64),
-    kernel_size=5)
+# multiple kernel_sizes of heatmap gaussian for 'Megvii' approach.
+kernel_sizes = [15, 11, 9, 7] + [11, 9, 7, 5]
```
ly015, line 34: Why it's the sum of 2 lists here with duplicated kernel size?

mmpose/datasets/transforms/formatting.py
liqikai9: Agreed. I have modified it.

configs/body_2d_keypoint/topdown_heatmap/coco/td-hm_2xrsn50_8xb32-210e_coco-256x192.py
liqikai9: This is because the encoding is prepared for 2 stages, and the kernel_sizes are the same for the last 3 levels in the first stage and the first 3 levels for the second stage. So I seperate the 2 stages' kernel_sizes to make it clear that each list of kernel sizes is for a stage.

On the one hand, I think the encoding may be redundant. But on the other hand, we should also leave the space for the possible modifications in configs. Is there any good solution to balance these two worries?

ly015: In this case, encoding with kernel sizes 7, 9, and 11 will be calculated twice which is unnecessary. I suggest user `kernel_sizes=[15, 11, 9, 7, 5]` and specify the level indices for different stages at the head configs.

mmpose/datasets/transforms/formatting.py
```diff
@@ -140,11 +140,21 @@ def transform(self, results: dict) -> dict:
...
+        gt_fields = None
         for key, packed_key in self.field_mapping_table.items():
             if key in results:
+                if isinstance(results[key], list):
+                    data_type = MultilevelPixelData
+                else:
+                    data_type = PixelData
+
+                if gt_fields is None:
+                    gt_fields = data_type()
```
ly015, line 152:
```suggestion
if isinstance(results[key], list):
    if gt_fields is None:
        gt_fields = MultilevelPixeldata()
    else:
        assert isinstance(gt_fields, MultilevelPixelData), 'Got mixed single-level and multi-level pixel data.'
else:
    # similar for PixelData
    ...

gt_fields.set_field(results[key], packed_key)     
```

ly015:
Please rebase onto the latest dev-1.x branch and fix conflicts. Also please check it can pass all unit tests.

liqikai9:
> Please rebase onto the latest dev-1.x branch and fix conflicts. Also please check it can pass all unit tests.

Done.

ly015 (approved):

PR #1654, opened 2022-09-14
[Refactor] Associative Embedding model

ly015:
## Motivation

Refactor Associative Embedding in 1.x

## Modification

In this PR:

- Add bottom-up pose estimator
- Add AE head and loss
- Add bottom-up image demo script
- Verified that the inference result is correct

To do in the future:

- Support bottom-up testing (need to modify the metric to work with bottom-up dataset)
- Support bottom-up training

## Checklist

**Before PR**:

**After PR**:


mmpose/codecs/simcc_label.py
```diff
@@ -105,15 +105,16 @@ def encode(
                 f'{self.smoothing_type}. Should be one of '
                 '{"gaussian", "standard"}')
 
-    def decode(self,
-               encoded: Tuple[np.ndarray,
-                              np.ndarray]) -> Tuple[np.ndarray, np.ndarray]:
+    def decode(self, simcc_x: np.ndarray,
```
ly015, line 108: @Tau-J In this PR I slightly changed the decoding process in the base head. Consequently, `codec.decode()` takes packed arguments instead of a single tuple. Is that ok with you?

Tau-J: No problem

configs/body_2d_keypoint/associative_embedding/coco/ae_hrnet-w32_8xb24-300e_coco-512x512.py
```diff
@@ -0,0 +1,155 @@
...
+            'pretrain_models/hrnet_w32-36af842e.pth'),
+    ),
+    head=dict(
+        type='AssociativeEmbeddingHead',
+        in_channels=32,
+        num_keypoints=17,
+        tag_dim=1,
+        tag_per_keypoint=True,
+        deconv_out_channels=None,
+        keypoint_loss=dict(type='KeypointMSELoss', use_target_weight=True),
```
jin-s13, line 91: why not use `combined loss` here?

ly015: AE uses exactly these 2 losses so I think it's better to explicitly include them in the argument list. CombinedLoss is more like a container (like ModuleDict), which is used in the `__init__()` to combine these 2 losses as one module (because the base head class assume there is one loss module).

demo/image_demo.py
```diff
@@ -25,7 +25,9 @@ def parse_args():
     return args
 
 
-def main(args):
+def main():
```
Ben-Louis, line 28: Is this demo necessary after the bottom-up models have been supported?

mmpose/apis/inference.py
```diff
@@ -176,14 +175,44 @@ def inference_topdown(model: nn.Module,
...
         results = []
+
+    return results
+
+
+def inference_bottomup(model: nn.Module, img: Union[np.ndarray, str]):
+    """Inference image with a bottom-up pose estimator.
+
+    Args:
+        model (nn.Module): The top-down pose estimator
```
Ben-Louis, line 195: 'top-down' -> 'bottom-up'

demo/image_demo.py
ly015: I think we can keep it for now in case someone wants to inference the model on cropped images.

Ben-Louis (approved):

PR #1662, opened 2022-09-15
[Refactor] Add Webcam API unittests

Ben-Louis:
## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/apis/webcam/utils/pose.py
+++ b/mmpose/apis/webcam/utils/pose.py
@@ -3,8 +3,8 @@ from typing import Dict, List, Tuple
 
 
 def get_eye_keypoint_ids(dataset_meta: Dict) -> Tuple[int, int]:
-    """A help function to get the keypoint indices of left and right eyes from
-    the dataset meta information.
+    """A helper function to get the keypoint indices of left and right eyes
+    from the dataset meta information.
 
     Args:
         dataset_meta (dict): dataset meta information.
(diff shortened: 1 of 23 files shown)


ly015:
#1638 has been merged. Please rebase the pr branch onto the latest dev-1.x

Ben-Louis:
The unittest for the missed lines, indeed the unittest for `webcam_executor.run`, involves multi-threading. 
It will cause `segmentation fault` in `pr_stage_test / build_cpu (3.7)` and fail while pass in other environments. Is there any method to fix this CI problem? @ly015

mmpose/apis/webcam/utils/pose.py
```diff
@@ -3,8 +3,8 @@
 
 
 def get_eye_keypoint_ids(dataset_meta: Dict) -> Tuple[int, int]:
-    """A helpfer function to get the keypoint indices of left and right eyes
-    from the dataset meta information.
+    """A help function to get the keypoint indices of left and right eyes from
```
ly015, line 6: I think the term 'helper function' is ok. Is there concern about it?

Ben-Louis: Actually not. I found the typo 'helpfer' and simply replaced it with 'help'. I'll change it now. 

BTW, maybe we also need to fix this type in master?

ly015 (approved):

PR #1668, opened 2022-09-19
[Fix] Fix a visualization bug and refine dataset browser

Ben-Louis:
## Modification

1. fix a bug about visualizer's dataset_meta.
      When testing with multiple GPUs, there are multiple processes but only one `visualizer`. The `visualizer.dataset_meta` is manually set to `None` occasionally in some processes in test_loop. This may lead to `AttributeError` in other processes when getting items from `visualizer.dataset_meta`.
3. refine dataset browser
    - use `file_client` to read image instead of `mmcv.imread`
    - keep `bbox_file` in dataset configs. Thus the visualization results match the images model receives indeed.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/tools/misc/browse_dataset.py
+++ b/tools/misc/browse_dataset.py
@@ -80,7 +80,8 @@ def main():
     cfg = Config.fromfile(args.config)
     if args.cfg_options is not None:
         cfg.merge_from_dict(args.cfg_options)
-    file_client = mmengine.FileClient(**cfg.file_client_args)
+    file_client_args = cfg.get('file_client_args', dict(backend='disk'))
+    file_client = mmengine.FileClient(**file_client_args)
 
     # register all modules in mmpose into the registries
     register_all_modules()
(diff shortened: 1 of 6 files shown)


Ben-Louis:
> `from __future__ import annotations` may have compatibility issues on python 3.6. Please consider using a string like 'BufferManager' for type hints within the class itself.

Solved. The corresponding unittest is add in [#1662](https://github.com/open-mmlab/mmpose/blob/ff644a287e624c0a3c9261f22e27a55a323fce43/tests/test_apis/test_webcam/test_utils/test_buffer.py#L73)

tools/misc/browse_dataset.py
```diff
@@ -80,6 +80,7 @@ def main():
     cfg = Config.fromfile(args.config)
     if args.cfg_options is not None:
         cfg.merge_from_dict(args.cfg_options)
+    file_client = mmengine.FileClient(**cfg.file_client_args)
```
ly015, line 83: We may need to check if `file_client_args` exists in the config.

Ben-Louis: Done.

ly015 (approved):

PR #1671, opened 2022-09-20
[Fix] Fix  keypoint_scores in visualization

liqikai9:
## Motivation

1. Support nms-thr in `topdown_demo_with_mmdet.py`;
2. Fix use `keypoint_scores` to filter keypoints to be visualized.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/demo/topdown_demo_with_mmdet.py
+++ b/demo/topdown_demo_with_mmdet.py
@@ -35,7 +35,7 @@ def visualize_img(args, img_path, detector, pose_estimator, visualizer,
         (pred_instance.bboxes, pred_instance.scores[:, None]), axis=1)
     bboxes = bboxes[np.logical_and(pred_instance.labels == args.det_cat_id,
                                    pred_instance.scores > args.bbox_thr)]
-    bboxes = bboxes[nms(bboxes, 0.3)][:, :4]
+    bboxes = bboxes[nms(bboxes, args.nms_thr)][:, :4]
 
     # predict keypoints
     register_mmpose_modules()
@@ -98,6 +98,11 @@ def main():
         type=float,
         default=0.3,
         help='Bounding box score threshold')
+    parser.add_argument(
+        '--nms-thr',
+        type=float,
+        default=0.3,
+        help=' Person detection score threshold for NMS')
     parser.add_argument(
         '--kpt-thr', type=float, default=0.3, help='Keypoint score threshold')
     parser.add_argument(
--- a/demo/topdown_demo_with_mmdet.py
+++ b/demo/topdown_demo_with_mmdet.py
@@ -102,7 +102,7 @@ def main():
         '--nms-thr',
         type=float,
         default=0.3,
-        help='Person detection score threshold for NMS')
+        help='IoU threshold for bounding box NMS')
     parser.add_argument(
         '--kpt-thr', type=float, default=0.3, help='Keypoint score threshold')
     parser.add_argument(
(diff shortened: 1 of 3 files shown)


mmpose/visualization/local_visualizer.py
```diff
@@ -225,8 +225,8 @@ def _draw_instances_kpts(self,
             keypoints = instances.get('transformed_keypoints',
                                       instances.keypoints)
 
-            if 'scores' in instances and self.show_keypoint_weight:
-                scores = instances.scores
+            if 'keypoint_scores' in instances:
```
Ben-Louis, line 228: Why is `self.show_keypoint_weight` omitted here?

demo/topdown_demo_with_mmdet.py
```diff
@@ -98,13 +98,18 @@ def main():
         type=float,
         default=0.3,
         help='Bounding box score threshold')
+    parser.add_argument(
+        '--nms-thr',
+        type=float,
+        default=0.3,
+        help='Person detection score threshold for NMS')
```
Ben-Louis, line 105:
```suggestion
        help='IoU threshold for bounding box NMS')
```

demo/topdown_face_demo.py
```diff
@@ -97,13 +97,18 @@ def main():
         'Default not saving the visualization images.')
     parser.add_argument(
         '--device', default='cuda:0', help='Device used for inference')
+    parser.add_argument(
+        '--nms-thr',
+        type=float,
+        default=0.3,
+        help='Face detection score threshold for NMS')
```
Ben-Louis, line 104:
```suggestion
        help='IoU threshold for bounding box NMS')
```

mmpose/visualization/local_visualizer.py
liqikai9: I think `keypoint_scores` shoud be irrevelant to `self.show_keypoint_weight`. 

A special case is that when I do not want to draw those keypoints with `keypoint_scores` lower than `kpt_score_thr`. If  `self.show_keypoint_weight = False`, and `keypoint_scores` is still present in instances, then although `keypoint_scores` is present in instances,  they are all forced to be assigned 1, and will be draw in the end.

Ben-Louis: Thanks, it makes sense. 

But `self.show_keypoint_weight` controls whether to plot keypoints with different transparency depending on keypoint_scores. When `self.show_keypoint_weight==False`, all visible keypoints should be plotted with the same transparency. This argument seems not to work now. To make it work, we also need to modify [this line](https://github.com/open-mmlab/mmpose/blob/e4549b13ea89dde824d5e4f90a7ef84c0364174d/mmpose/visualization/local_visualizer.py#L262).

liqikai9: How about this
```
if self.show_keypoint_weight:
    transparency = max(0, min(1, score[kid])) * self.alpha
```

Ben-Louis: `transparency` needs to be set when `self.show_keypoint_weight` is False
```python
transparency = self.alpha
if self.show_keypoint_weight:
    transparency *= max(0, min(1, score[kid]))  
```

liqikai9: Nice. I will fix it according to your advice.

Ben-Louis (approved):

PR #1693, opened 2022-09-28
[Feature] support DEKR

JinluZhang1126:
## Motivation

To support the single-stage model, DEKR, which is missed in previous versions of MMPose.

## Modification

add single-stage config files in the './configs' directory;
add OneStage detector class and corresponding regression head in the './mmpose/model' directory;
update the bottom-up data transformation like generating offsets, multi-scale, and match_to_heatmap in the './mmpose/datasets' directory...

## BC-breaking (Optional)

No.

## Use cases (Optional)
```bash
python tools/train.py configs/body/2d_kpt_sview_rgb_img/onestage/coco/hrnet_w48_coco_640x640.py
```

## Checklist

**Before PR**:

**After PR**:


configs/body/2d_kpt_sview_rgb_img/disentangled_keypoint_regression/README.md
```diff
@@ -0,0 +1,22 @@
+# Bottom-up Human Pose Estimation via Disentangled Keypoint Regression (DEKR)
```
ly015, line 1: I suggest using 'dekr' as the folder name for simplicity.

mmpose/datasets/pipelines/bottom_up_transform.py
```diff
@@ -733,41 +976,13 @@ def __call__(self, results):
             input_size = np.array([input_size, input_size], dtype=np.int)
         img = results['img']
 
-        h, w, _ = img.shape
-
-        # calculate the size for min_scale
-        min_input_w = _ceil_to_multiples_of(self.min_scale * input_size[0], 64)
```
jin-s13, line 739: please double check this part.

mmpose/datasets/pipelines/bottom_up_transform.py
```diff
@@ -572,38 +708,138 @@ def __call__(self, results):
...
+                                    '`gen_center_heatmap` is True. ' \
+                                    'e.g. sigma=(2, 4)'
+
         self.sigma = sigma
+        self.bg_weight = bg_weight
+        self.gen_center_heatmap = gen_center_heatmap
         self.use_udp = use_udp
 
-    def _generate(self, num_joints, heatmap_size):
+    def _generate(self, num_joints, sigma, heatmap_size):
```
jin-s13, line 745: is sigma necessary?

Ben-Louis: `BottomUpGenerateHeatmapTarget` is modified to support the generation of both keypoint and instance center heatmaps. These two kinds of heatmaps are generally rendered with different sigma. 

Thus, `self.sigma` is changed to a tuple that contains sigmas for keypoint and instance center heatmaps. The parameter `sigma` of function `_generate` is the sigma for either keypoint or instance center heatmaps

mmpose/datasets/pipelines/bottom_up_transform.py
Ben-Louis: Thanks, the option `use_udp` is lost. After fixing this problem, I compare the deleted codes with the function `_get_multi_scale_size` and find that they match well. On 10000 random inputs, two versions of `BottomUpGetImgSize` generate the same outputs

ly015 (approved):

PR #1771, opened 2022-10-26
Fix broken links

doufengqi:
## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/docs/en/tasks/2d_animal_keypoint.md
+++ b/docs/en/tasks/2d_animal_keypoint.md
@@ -35,7 +35,7 @@ MMPose supported datasets:
 
 For [Animal-Pose](https://sites.google.com/view/animal-pose/) dataset, we prepare the dataset as follows:
 
-1. Download the images of [PASCAL VOC2012](http://host.robots.ox.ac.uk/pascal/VOC/voc2012/#data), especially the five categories (dog, cat, sheep, cow, horse), which we use as trainval dataset.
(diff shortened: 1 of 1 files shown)


doufengqi:
The previous link cannot be opened (not an isolated case), here is the repair link

jin-s13:
I tried the previous link in the (md file)[https://github.com/open-mmlab/mmpose/blob/master/docs/en/tasks/2d_animal_keypoint.md] just now, but it worked fine...

ly015:
@doufengqi Thank you very much for your help. I think the original link is not broken. It links to a section of the official site, instead of the download URL. I think using the download URL here could be more clear and more convenient for users.

doufengqi:
> @doufengqi Thank you very much for your help. I think the original link is not broken. It links to a section of the official site, instead of the download URL. I think using the download URL here could be more clear and more convenient for users.

We tried many times to visit yesterday and failed, but today it suddenly became better. However, my new link can directly download the target file, and people don't need to click in to find the download link.

doufengqi:
> I tried the previous link in the [md file](https://github.com/open-mmlab/mmpose/blob/master/docs/en/tasks/2d_animal_keypoint.md) just now, but it worked fine...

We tried many times to visit yesterday and failed, but today it suddenly became better. However, my new link can directly download the target file, and people don't need to click in to find the download link.

PR #1834, opened 2022-11-24
[Refactor] Support DEKR

Ben-Louis:
## Checklist

**Before PR**:

**After PR**:


mmpose/evaluation/functional/nms.py
```diff
@@ -249,3 +249,75 @@ def soft_oks_nms(kpts_db: List[dict],
...
+
+
+def nearby_joints_nms(
+    kpts_db,
+    dist_thr,
+    num_nearby_joints_thr=None,
+    score_per_joint=False,
+    max_dets: int = 30,
+):
+    """Nearby joints NMS implementations.
```
ly015, line 261: Could you please add a description or paper reference of the algorithm?

ly015: And it would be better to add type hints as well.

mmpose/models/heads/regression_heads/dekr_head.py
```diff
@@ -0,0 +1,605 @@
+# Copyright (c) OpenMMLab. All rights reserved.
```
ly015, line 1: Not sure that the DEKR head belongs to regression heads because it predict roots and offsets in a heatmap-based manner. Maybe it's more like a heatmap head or should be put into a new category like 'single-stage head' or 'hybrid head'?

Ben-Louis: The localization of instance roots is based on heatmaps but the keypoint coordinates are calculated via regressed offset. In my perspective, the dense **offset** map is dissimilar from heatmap whose **activation** plays a key role. 

Maybe 'hybrid head' is a better choice than 'single-stage head' because some single-stage methods such as CID use only heatmap. It might be hard to categorize them if the latter is adopted.

mmpose/evaluation/functional/nms.py
Ben-Louis: This function is modified from the implementation in the official code. The reference to their repo has been added

ly015 (approved):

PR #1882, opened 2022-12-14
[Refactor] Align checkpoint intervals with master

Ben-Louis:
## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/apis/webcam/nodes/model_nodes/detector_node.py
+++ b/mmpose/apis/webcam/nodes/model_nodes/detector_node.py
@@ -124,7 +124,7 @@ class DetectorNode(Node):
         """Post-process the predictions of MMDetection model."""
         instances = preds.pred_instances.cpu().numpy()
 
-        classes = self.model.dataset_meta['CLASSES']
+        classes = self.model.dataset_meta['classes']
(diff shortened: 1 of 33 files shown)


mmpose/apis/webcam/nodes/model_nodes/detector_node.py
```diff
@@ -124,7 +124,7 @@ def _post_process(self, preds) -> List[Dict]:
         """Post-process the predictions of MMDetection model."""
         instances = preds.pred_instances.cpu().numpy()
 
-        classes = self.model.dataset_meta['CLASSES']
+        classes = self.model.dataset_meta['classes']
```
ly015, line 127: Is the field name updated in a recent mmdet version? Do we need to consider compatibility to both 'CLASSES' and 'classes'?

Ben-Louis: Yes, MMDetection has changed all their metainfo keys from upper case to lower case. This modification does not lead to incompatibility in MMPose with metainfo key 'CLASSES' which is only used in metrics.

ly015 (approved):

PR #1900, opened 2022-12-27
[Fix] replace np.int with int

jin-s13:
## Motivation

numpy 1.24.0 removes np.bool, np.float, np.int. 
To fix it, I replace np.int (and some np.int64) with np.int32.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/datasets/datasets/body3d/body3d_mview_direct_panoptic_dataset.py
+++ b/mmpose/datasets/datasets/body3d/body3d_mview_direct_panoptic_dataset.py
@@ -211,7 +211,7 @@ class Body3DMviewDirectPanopticDataset(Kpt3dMviewRgbImgDirectDataset):
                             dtype=np.float32)
 
                         cnt = 0
-                        person_ids = -np.ones(self.max_persons, dtype=np.int32)
+                        person_ids = -np.ones(self.max_persons, dtype=int)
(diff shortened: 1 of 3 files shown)


jin-s13:
Same problems exist in xtcocotools, I will fix xtcocotools. 
We may need to upgrade xtcocotools afterwards.

mmpose/datasets/datasets/body3d/body3d_mview_direct_panoptic_dataset.py
```diff
@@ -211,7 +211,7 @@ def _get_db(self):
                             dtype=np.float32)
 
                         cnt = 0
-                        person_ids = -np.ones(self.max_persons, dtype=np.int)
```
ly015, line 214: `np.int` was an alias to `int` and would be int64 in a 64-bit system. If so, should we keep this behavior?

jin-s13: ok. let's use int instead.

ly015 (approved):

dancingdarwin:
Hello, I'd like to kindly follow up on this pull request. I tried downgrading numpy to 1.20.0 but then I get other errors from xtocotools. Thanks so much!

PR #1907, opened 2022-12-30
[Refactor] Support CID

Ben-Louis:
## Modification

1. add CIDHead
2. add a codec for CID named 'DecoupledHeatmap', which means that the heatmaps for each instance are decoupled and represented on different channels
3. add configs

TODO:
- [x] add unittests 
- [x] add docstrings

## Checklist

**Before PR**:

**After PR**:


mmpose/models/heads/heatmap_heads/cid_head.py
```diff
@@ -0,0 +1,727 @@
...
+                instance_heatmaps_flip = \
+                    instance_heatmaps_flip[:, flip_indices, :, :]
+                instance_heatmaps = (instance_heatmaps +
+                                     instance_heatmaps_flip) / 2.0
+            instance_heatmaps = smooth_heatmaps(
+                instance_heatmaps, test_cfg.get('blur_kernel_size', 3))
+
+            preds = self.decode((instance_heatmaps, instance_scores[:, None]))
+            preds = InstanceData.cat(preds)
+            preds.keypoints[..., :2] += 1.5
```
jin-s13, line 581: why 1.5?

mmpose/models/heads/heatmap_heads/cid_head.py
```diff
@@ -0,0 +1,727 @@
...
+
+        self.channel_attention = ChannelAttention(in_channels, gfd_channels)
+        self.spatial_attention = SpatialAttention(in_channels, gfd_channels)
+        self.fuse_attention = build_conv_layer(
+            dict(
+                type='Conv2d',
+                in_channels=gfd_channels * 2,
+                out_channels=gfd_channels,
+                kernel_size=1))
+        self.heatmap_conv = build_conv_layer(
```
jin-s13, line 334: In mmpose 0.x, heatmap_conv.bias and keypoint_center_conv.bias is initialized as bias_value. Do we still have this feature in 1.x?

bias_value = -math.log((1 - self.prior_prob) / self.prior_prob)
torch.nn.init.constant_(self.keypoint_center_conv.bias, bias_value)
torch.nn.init.constant_(self.heatmap_conv.bias, bias_value)

mmpose/models/heads/heatmap_heads/cid_head.py
Ben-Louis: For example, consider x. Its range in the input image space is [0, w-1], but in the heatmap space, it is [0, w-4] after being rescaled. The shift is (4-1)/2=1.5. 

However, I have found that using `preds.keypoints[..., :2] += 2` results in higher accuracy. I will fix this in a later commit. Do you have any idea about the reason?

mmpose/models/heads/heatmap_heads/cid_head.py
Ben-Louis: This initialization has not been added yet because the alignment of training accuracy has not been started. Thanks for your reminder

mmpose/models/heads/heatmap_heads/cid_head.py
jin-s13: 0.5 * 4 (down-sampling scale) = 2?

mmpose/models/losses/classification_loss.py
```diff
@@ -153,3 +153,41 @@ def forward(self, pred_simcc, gt_simcc, target_weight):
                 self.criterion(coord_y_pred, coord_y_gt).mul(weight).sum())
 
         return loss / num_joints
+
+
+@MODELS.register_module()
+class InfoNCELoss(nn.Module):
+    """InfoNCE loss for training a discriminative representation space with a
+    contrastive manner.
```
ly015, line 161: Could you please add the paper reference for InforNCELoss and FocalHeatmapLoss?

Ben-Louis: Sure. The references have been added

ly015 (approved):

PR #1944, opened 2023-01-28
[Feature] Keypoint partition metric

zwfcrazy:
## Motivation

Sometimes one may be interested in the performance of a pose model on certain body parts rather than on all the keypoints. For example, ``CocoWholeBodyMetric`` evaluates coco metric on body, foot, face, lefthand and righthand. However, ``CocoWholeBodyMetric`` is not flexible enough to be applied to arbitrary custom datasets. Therefore, we provide this wrapper metric.

## Modification

A new wrapper metric class is added.
Unit tests have been added.

## BC-breaking (Optional)

No.

## Use cases (Optional)

In my dataset, there are 40 keypoints. Some belongs to the body (coco 17 points), some belongs to foot and hands, etc. In the config file, we can do the following
```python
val_evaluator = dict(
        type='KeypointPartitionMetric',
        metric=dict(
            type='CocoMetric',
        ),
        partitions=dict(
            body=range(17),
            foot=range(17, 23),
            jaw_lip=range(23, 28),
            left_hand=range(28, 34),
            right_hand=range(34, 40),
            all=range(40)
        )
)
where the numbers of each partition are keypoint indices. Note, the indices can be discontinuous.
```

## Checklist

**Before PR**:

**After PR**:


mmpose/evaluation/metrics/keypoint_partition_metric.py
```diff
@@ -0,0 +1,132 @@
...
+    enough to be applied to arbitrary custom datasets. Therefore, we provide
+    this wrapper metric.
+
+    Supported metrics:
+        ``CocoMetric``  all keypoint ground truth should be stored in
+            `keypoints` not other data fields.
+        ``PCKAccuracy`` the required data fields should be provided in
+            annotation files, such as head bbox, etc.
+        ``AUC``
+        ``EPE``
```
Ben-Louis, line 30: The descriptions of supported metrics are inadequate; BTW, it would be beneficial to include a type check on the wrapped metric in the `__init__` function.

mmpose/evaluation/metrics/keypoint_partition_metric.py
```diff
@@ -0,0 +1,132 @@
...
+        # instantiate metrics for each partition
+        self.metrics = {
+            partition_name: METRICS.build(metric)
+            for partition_name in partitions.keys()
+        }
+
+    @property
+    def dataset_meta(self) -> Optional[dict]:
+        """Optional[dict]: Meta info of the dataset."""
+        return self._dataset_meta
```
Ben-Louis, line 64: This function seems to be the same as the base class. Is it necessary?

mmpose/evaluation/metrics/keypoint_partition_metric.py
```diff
@@ -0,0 +1,132 @@
...
+
+        for partition_name, metric in self.metrics.items():
+            metric.process(data_batch, parted_data_samples[partition_name])
+
+    def compute_metrics(self, results: list) -> dict:
+        pass
+
+    def evaluate(self, size: int) -> dict:
+        """Overload the evaluate method of the base class.
+
```
Ben-Louis, line 122: I think this line of comment can be removed for the sake of simplicity. What do you think?

mmpose/evaluation/metrics/keypoint_partition_metric.py
```diff
@@ -0,0 +1,132 @@
...
+                _data_sample['pred_instances']['keypoints'] = _data_sample[
+                    'pred_instances']['keypoints'][:, keypoint_ids]
+                _data_sample['gt_instances']['keypoints'] = _data_sample[
+                    'gt_instances']['keypoints'][:, keypoint_ids]
+                _data_sample['gt_instances'][
+                    'keypoints_visible'] = _data_sample['gt_instances'][
+                        'keypoints_visible'][:, keypoint_ids]
+
+                # for coco metric
+                if 'raw_ann_info' in _data_sample:
```
Ben-Louis, line 100: Is this wrapper only compatible with CocoMetric instantiated without `ann_file`?

mmpose/evaluation/metrics/keypoint_partition_metric.py
zwfcrazy: I think AUC and EPE can be used without any condition, so I didn't write anything there. It is a good idea to do a type check.

mmpose/evaluation/metrics/keypoint_partition_metric.py
zwfcrazy: The setter and getter is a pair, if it is removed, an error will be thrown.

mmpose/evaluation/metrics/keypoint_partition_metric.py
zwfcrazy: Yes, I forgot to mention this.

mmpose/evaluation/metrics/keypoint_partition_metric.py
zwfcrazy: I think the evaluate method was not designed to be overloaded, so I mention it here in case someone don't know. Anyway, of course it can be removed.

mmpose/evaluation/metrics/keypoint_partition_metric.py
Tau-J: @Ben-Louis I think this partition metric is mainly designed for custom/industrial applications, so it could be acceptable. @ly015 What do you think?

Ben-Louis: I agree. I'm considering the situation where users provide an `ann_file`. Would it be better to raise an AssertionError or simply neglect it with a warning? @Tau-J

mmpose/evaluation/metrics/keypoint_partition_metric.py
Ben-Louis: Thanks

zwfcrazy: I have found a way to implement this.

Tau-J (approved):

Ben-Louis (approved):

PR #1960, opened 2023-02-07
[Fix] Associative Embedding inference align with master

ly015:
## Checklist

**Before PR**:

**After PR**:


mmpose/datasets/transforms/bottomup_transforms.py
```diff
@@ -478,6 +478,7 @@ def transform(self, results: Dict) -> Optional[dict]:
                     output_size=actual_input_size)
             else:
                 center = np.array([img_w / 2, img_h / 2], dtype=np.float32)
+                center = np.round(center)
```
ly015, line 481: @Ben-Louis  This rounding is implemented in the [master branch](https://github.com/open-mmlab/mmpose/blob/master/mmpose/datasets/pipelines/bottom_up_transform.py#L76), but missing in the 1.x branch. It will affect the performance of AE models. Could you please check its influence on DEKR?

Ben-Louis: > @Ben-Louis This rounding is implemented in the [master branch](https://github.com/open-mmlab/mmpose/blob/master/mmpose/datasets/pipelines/bottom_up_transform.py#L76), but missing in the 1.x branch. It will affect the performance of AE models. Could you please check its influence on DEKR?

For DEKR models, rounding the center can result in a reduction of about 0.1 AP on COCO

PR #2039, opened 2023-03-10
[Fix] Avoid scope switching when using mmdet inference interface

Ben-Louis:
## Motivation

While inferring top-down models on videos, the scope alternates between 'mmdet' and 'mmpose' repeatedly, which often results in numerous warning messages.
屏幕截图 2023-03-10 154340

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/utils/config_utils.py
+++ b/mmpose/utils/config_utils.py
@@ -2,7 +2,7 @@
 from mmpose.utils.typing import ConfigDict
 
 
-def convert_mmdet_test_pipeline(cfg: ConfigDict) -> ConfigDict:
+def adapt_mmdet_pipeline(cfg: ConfigDict) -> ConfigDict:
     """Converts pipeline types in MMDetection's test dataloader to use the
     'mmdet' namespace.
 
--- a/mmpose/utils/config_utils.py
+++ b/mmpose/utils/config_utils.py
@@ -1,0 +1,16 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from mmpose.utils.typing import ConfigDict
+
+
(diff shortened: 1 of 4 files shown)


mmpose/utils/config_utils.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from mmpose.utils.typing import ConfigDict
+
+
+def convert_mmdet_test_pipeline(cfg: ConfigDict) -> ConfigDict:
```
ly015, line 5: What about naming the function 'port_mmdet_pipeline' or 'adapt_mmdet_pipeline'?

ly015:
BTW, please also use lazy import for mmdetection in other parts of mmpose package, like `https://github.com/open-mmlab/mmpose/blob/dev-1.x/mmpose/apis/inferencers/pose2d_inferencer.py#L7`. So the model training/testing and document compiling will not depend on mmdetection.

ly015:
> BTW, please also use lazy import for mmdetection in other parts of mmpose package, like `https://github.com/open-mmlab/mmpose/blob/dev-1.x/mmpose/apis/inferencers/pose2d_inferencer.py#L7`. So the model training/testing and document compiling will not depend on mmdetection.

Fixed in #2045 . No need to update this PR.

mmpose/utils/config_utils.py
Ben-Louis: Good idea. I think 'adapt_mmdet_pipeline' is preferable

ly015 (approved):

PR #2130, opened 2023-03-28
[Fix] fix simcc visualization

Tau-J:
## Checklist

**Before PR**:

**After PR**:


Ben-Louis:
Is there a particular reason for calling `get_simcc_normalized` within `head.predict` rather than using it in `codec.decode`?

Tau-J:
After testing, simcc's accuracy decreased by 0.6 AP using `get_simcc_normalized()`. RTMPose remained unchanged. So I decide to call this method only when performing heatmap visualization

Ben-Louis:
The code looks good to me. But I'm concerned about the discrepancy between the keypoint scores and the 1D heatmaps generated by SimCC, as this could cause confusion. Do you think it would be beneficial to include clarifying instructions somewhere?

Ben-Louis (approved):

PR #2139, opened 2023-04-01
[Feature] Support AnimalKingdom dataset

Dominic23331:
Add AnimalKingdom dataset


Ben-Louis:
Thank you so much for your valuable contribution! By the way, have you tried to train any model on this dataset?

Dominic23331:
Not yet. I'll try it in a few days。

Ben-Louis:
> Not yet. I'll try it in a few days。

Thank you. We shall anticipate the inclusion of the model configuration files. Moreover, it is suggested to incorporate a concise overview of this dataset, together with the file hierarchy, in the `docs/en/dataset_zoo/2d_animal_keypoint.md`.

Dominic23331:
I trained AK_ P1 and AK_ P2 two datasets, but the effect is not very good, which is more than three points less than the results in the paper

Ben-Louis:
> I trained AK_ P1 and AK_ P2 two datasets, but the effect is not very good, which is more than three points less than the results in the paper

Would you mind sharing your config and training log? That way, we can help figure out what's going wrong.

Dominic23331:
> > I trained AK_ P1 and AK_ P2 two datasets, but the effect is not very good, which is more than three points less than the results in the paper
> 
> Would you mind sharing your config and training log? That way, we can help figure out what's going wrong.

Okay, I will upload the config file as soon as possible in the next two days

Dominic23331:
I upload config files and added explanatory documents

Ben-Louis:
> I upload config files and added explanatory documents

Thanks. Have you trained models with any of them in order to verify their accuracy?

Dominic23331:
> > I upload config files and added explanatory documents
> 
> Thanks. Have you trained models with any of them in order to verify their accuracy?

I have tried to train several classes, but the results are not very satisfactory. I will consider how to improve it.

Dominic23331:
Due to the busy work ahead, I apologize for the delay in this PR. During testing, I found that 210 epochs were not enough to fit the training set, so I increased the epochs to 300.

Dominic23331:
This is the result of my training
                    paper    mmpose
ALL              66.06     62.72
leave-k-out 39.3       37.74
mammals    61.59     57.56
amphibians 56.74     53.56
reptiles        56.06     50
birds            77.35     76.79
fishes           68.25     64.3
This result is not ideal

Tau-J (approved):

PR #2108, opened 2023-03-22
[Refactor] Add FeatureMapProcessor to replace transform_inputs in heads

Ben-Louis:
## Modification

1. Introduced a new neck, `FeatureMapProcessor`, which is capable of selecting, concatenating, and rescaling feature maps generated by backbone layers.
2. Removed the `_transform_inputs` and `_get_in_channels` functions from heads.
3. Updated configs according to the above modifications. 

The test accuracy of several representative models affected by these changes has been verified:
- `cid_hrnet-w32_8xb20-140e_coco-512x512`
- `dekr_hrnet-w32_8xb10-140e_coco-512x512`
- `dekr_hrnet-w32_8xb10-300e_crowdpose-512x512`
- `ipr_res50_8xb64-210e_coco-256x256`
- `rtmpose-t_8xb256-420e_coco-256x192`
- `simcc_res50_8xb64-210e_coco-256x192`
- `td-hm_hrnet-w32_8xb64-210e_coco-256x192`
- `td-hm_pvt-s_8xb64-210e_coco-256x192`
- `td-hm_ViTPose-base-simple_8xb64-210e_coco-256x192`
- `td-reg_res50_8xb64-210e_coco-256x192`
- `td-reg_res50_rle-pretrained-8xb64-210e_coco-256x192`

## Checklist

**Before PR**:

- [ ] I have read and followed the workflow indicated in the [CONTRIBUTING.md](https://github.com/open-mmlab/mmpose/blob/master/.github/CONTRIBUTING.md) to create this PR.
- [ ] Pre-commit or linting tools indicated in [CONTRIBUTING.md](https://github.com/open-mmlab/mmpose/blob/master/.github/CONTRIBUTING.md) are used to fix the potential lint issues.
- [ ] Bug fixes are covered by unit tests, the case that causes the bug should be added in the unit tests.
- [ ] New functionalities are covered by complete unit tests. If not, please add more unit tests to ensure correctness.
- [ ] The documentation has been modified accordingly, including docstring or example tutorials.

**After PR**:

- [ ] CLA has been signed and all committers have signed the CLA in this PR.

PR #2153, opened 2023-04-04
[Refactor] add config checker

Ben-Louis:
## Motivation

 #2108 and #2126 modified the interface of neck and head. If users use the latest mmpose and old configs, there might be some conflict.

## Modification

1. add a function `check_and_update_config` that checks if the config is consistent with the latest mmpose. If not, the arguments will be automatically updated. In addition, this function will display the guide for updating the configs, such as 
image
and
image

2. add unit test for  `check_and_update_config`
3. verify the test accuracy of `cid_hrnet-w32_8xb20-140e_coco-512x512`, `td-hm_ViTPose-base-simple_8xb64-210e_coco-256x192`, `td-hm_pvt-s_8xb64-210e_coco-256x192`

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/models/pose_estimators/base.py
+++ b/mmpose/models/pose_estimators/base.py
@@ -7,6 +7,7 @@ from mmengine.model import BaseModel
 from torch import Tensor
 
 from mmpose.datasets.datasets.utils import parse_pose_metainfo
+from mmpose.models.utils import check_and_update_config
 from mmpose.registry import MODELS
 from mmpose.utils.typing import (ConfigType, ForwardResults, OptConfigType,
                                  Optional, OptMultiConfig, OptSampleList,
@@ -44,6 +45,12 @@ class BasePoseEstimator(BaseModel, metaclass=ABCMeta):
         self.metainfo = self._load_metainfo(metainfo)
 
         self.backbone = MODELS.build(backbone)
+
+        # the PR #2108 and #2126 modifies the interface of neck and head.
+        # The following function automatically detects outdated
+        # configurations and updates them accordingly, while also providing
+        # clear and concise information on the changes made.
+        neck, head = check_and_update_config(neck, head)
 
         if neck is not None:
             self.neck = MODELS.build(neck)
--- a/mmpose/models/pose_estimators/base.py
+++ b/mmpose/models/pose_estimators/base.py
@@ -46,7 +46,7 @@ class BasePoseEstimator(BaseModel, metaclass=ABCMeta):
 
         self.backbone = MODELS.build(backbone)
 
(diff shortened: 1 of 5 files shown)


mmpose/models/utils/check_and_update_config.py
```diff
@@ -0,0 +1,230 @@
...
+                 'The following section of the config:\n```\n'
+    total_info += old_model_info
+    total_info += '```\nshould be updated to\n```\n'
+    total_info += new_model_info
+    total_info += '```\nFor more information, please refer to '\
+                  'https://mmpose.readthedocs.io/en/dev-1.x/' \
+                  'guide_to_framework.html#step3-model'
+
+    logger: MMLogger = MMLogger.get_current_instance()
+    logger.info(total_info)
```
Tau-J, line 173: In my opinion, this info should be a user `warning` instead of  `info`

mmpose/models/pose_estimators/base.py
```diff
@@ -45,6 +46,12 @@ def __init__(self,
 
         self.backbone = MODELS.build(backbone)
 
+        # the PR #2108 and #2126 modified the interface of neck and head.
+        # The following function automatically detects outdated
+        # configurations and updates them accordingly, while also providing
+        # clear and concise information on the changes made.
+        neck, head = check_and_update_config(neck, head)
```
Tau-J, line 53: I think this function should also take the backbone cfg as input, so that it can be used as a more general check and auto-update process for the entire model, rather than just for compatibility measures with the previous PR.

Tau-J: When trying to inference a model with `SyncBN` on `cpu`, it will raise an error. I think we can check the device and automatically repalce `SyncBN` as `BN` in the config dict.

Ben-Louis: > When trying to inference a model with `SyncBN` on `cpu`, it will raise an error. I think we can check the device and automatically repalce `SyncBN` as `BN` in the config dict.

`SyncBN` can be converted to `BN` by `mmengine.model.revert_sync_batchnorm`. The Inferencer has supported this conversion in @2152

Ben-Louis: > I also have a concern that at the beginning of the training, `mmengine` will print the entire config and dump a config file. Will this automatically processed function affect the above content? Such automatic modification seems to cause differences between the real model structure and the output. Do you think there is any way to optimize it?

We can modify this checker to a hook in the future, and add more functions to it.

Tau-J (approved):

PR #2237, opened 2023-04-16
[MMSIG-80] Update and refine get_flops.py

xin-li-67:
## Motivation

MMSIG-80: update and refine get_flops.py

## Modification

tools/analysis_tools/get_flops.py

## Checklist

**Before PR**:

**After PR**:


tools/analysis_tools/get_flops.py
```diff
@@ -29,13 +32,13 @@ def parse_args():
...
         type=int,
         nargs='+',
         default=[256, 192],
         help='input image size')
     parser.add_argument(
-        '--input-constructor',
+        '--batch-input',
         '-c',
         type=str,
         choices=['none', 'batch'],
```
Ben-Louis, line 44: Perhaps the argument can be converted to a boolean type using the `store_true` action? Additionally, the option '-c' may no longer be intuitive, as the argument name has been altered.

xin-li-67: > Perhaps the argument can be converted to a boolean type using the `store_true` action? Additionally, the option '-c' may no longer be intuitive, as the argument name has been altered.

agree! I will update this in the next commit

Tau-J:
1. `--batch-input` and `--batch-size` seems to be cumbersome. How about use `batch_input = args.batch_size > 1`?
2. I found a bug in mmengine: when `inputs=None`, `get_model_complexity_info()` cannot inference with `cuda:0`. I suggest set `cpu`  as default device for the moment. I'll report this bug to mmengine, or maybe someone will raise a PR to fix it.
3. The message is too long when printing both `out_arch` and `out_table`. I suggest only printing `out_table` and leaving `out_arch` as an optional argument.

xin-li-67:
> 1. `--batch-input` and `--batch-size` seems to be cumbersome. How about use `batch_input = args.batch_size > 1`?
> 2. I found a bug in mmengine: when `inputs=None`, `get_model_complexity_info()` cannot inference with `cuda:0`. I suggest setting `cpu`  as default device for the moment. I'll report this bug to mmengine, or maybe someone will raise a PR to fix it.
> 3. The message is too long when printing both `out_arch` and `out_table`. I suggest only printing `out_table` and leaving `out_arch` as an optional argument.

I made some changes in the latest commit:
1. removed `batch-input` and use `batch-size`;
2. now default device is CPU;
3. now `out-table` will be printed by default, and `out-arch` will not be printed unless the user specifies with `--show-arch-info`. However, it seems that the `get_complexity_info` function in MMEngine still needs to be refined, some models in MMPose are not well supported.

Tau-J:
LGTM

Tau-J (approved):

PR #2244, opened 2023-04-17
[Enhance] Support Webcam inference in topdown/bottomup demo.

Tau-J:
## Motivation

- Support Webcam inference in topdown/bottomup demo.
- Support FastVisualizer to conduct efficient demo inference.

## Checklist

**Before PR**:

**After PR**:


mmpose/visualization/fast_visualizer.py
```diff
@@ -0,0 +1,61 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import cv2
+
+
+class FastVisualizer:
```
Ben-Louis, line 5: It would be better if we added a unit test for `FastVisualizer`

Tau-J: You're right. I'll add a ut later.

Tau-J:
> Do you think it is necessary to ensure that the interfaces of `FastVisualizer` remain consistent with those of `PoseLocalVisualizer`, similar to the visualizers in other OpenMMLab repos?

That's necessary. But how about leaving this kind of stuff as the community assignment?

Ben-Louis:
> That's necessary. But how about leaving this kind of stuff as the community assignment?

Sure, no problem

demo/bottomup_demo.py
```diff
@@ -81,13 +85,15 @@ def parse_args():
...
         type=int,
-        default=3,
+        default=6,
         help='Keypoint radius for visualization')
     parser.add_argument(
         '--thickness',
         type=int,
-        default=1,
+        default=3,
         help='Link thickness for visualization')
```
Ben-Louis, line 94: Would these two values be excessively large for visualization using `PoseLocalVisualizer`, which is adopted for image inputs?

Ben-Louis (approved):

PR #2334, opened 2023-05-05
[Fix] config param and return values of functions in the dataset class in the tutorial

shuheilocale:
## Motivation

MMPose_Tutorial.ipynb output an error at runtime. Therefore, I fixed it.

## Modification

Two modifications are as follows

- Return value of _load_annotations()
  - The caller of the _load_annotations function expects two values to be returned, but the code before the modification returned only one value. Therefore, this was changed to return two values. However, the second return value is not used by the caller, so None is returned.
- Metrics
  - There was no "pck/PCK@0.0" metric, so I corrected it to "PCK".

## Checklist

**Before PR**:

**After PR**:


Tau-J:
@shuheilocale Hi, would you mind signning the CLA? I'll fix the ci of docs

shuheilocale:
@Tau-J 

Sorry, I agreed CLA, but it seems that the email address on the commit was unusual, and when I ran the command to fix it, the commit was wrong.
So I'm going to withdraw this PR and PR again.

PR #2386, opened 2023-05-22
[Feature] add visibility prediction head

Billccx:
## Motivation
Enable top-down methods to predict keypoints visibility

## Modification
1. add a new head `vis_head`      (mmpose/mmpose/models/heads/vis_heads)
2. add a new TTA function `flip_visibility`     (mmpose/mmpose/models/utils/tta.py)
3. add a new key `keypoints_visible` in `label_mapping_table` (mmpose/mmpose/datasets/transforms/formatting.py)

## Use cases (Optional)

We can use `vis_head` to wrap a keypoints prediction head, getting the keypoints prediction results with keypoints visibility.

For example, in config file:

```
head=dict(
      type='VisibilityPredictionHead',
      in_channels=2048,
      num_joints=17,
      loss=dict(type='BCELoss', use_target_weight=True),
      pose_cfg=dict(
          type='RegressionHead',
          in_channels=2048,
          num_joints=17,
          loss=dict(type='SmoothL1Loss', use_target_weight=True),
          decoder=codec
      )
  )
```

## Checklist

**Before PR**:

**After PR**:


mmpose/models/heads/vis_heads/vis_head.py
```diff
@@ -0,0 +1,213 @@
...
+            feats (Tuple[Tensor]): Multi scale feature maps.
+
+        Returns:
+            Tensor: output coordinates visibility.
+        """
+        x = feats[-1]
+        while len(x.shape) < 4:
+            x.unsqueeze_(-1)
+        x = self.vis_head(x)
+        return x.reshape(-1, self.num_joints)
```
Tau-J, line 66: This implementation will cause the `pose_head` to be lost during the export of onnx. Also, the user must manually call `pose_head` for inference, instead of getting all results at once, which is not ideal.
I suggest performing `pose_head` inference in `forward()` and returning results for both `vis_head` and `pose_head` at the same time.

mmpose/models/heads/__init__.py
```diff
@@ -6,9 +6,11 @@
 from .hybrid_heads import DEKRHead
 from .regression_heads import (DSNTHead, IntegralRegressionHead,
                                RegressionHead, RLEHead)
+from .vis_heads import VisibilityPredictionHead
```
Tau-J, line 9: Head with multiple outputs can be put in `hybrid_heads` folder.

mmpose/models/heads/vis_heads/vis_head.py
```diff
@@ -0,0 +1,213 @@
...
+        self.in_channels = in_channels
+        self.num_joints = num_joints
+        self.loss_module = MODELS.build(loss)
+
+        self.pose_head = MODELS.build(pose_cfg)
+        self.pose_cfg = pose_cfg
+
+        self.vis_head = nn.Sequential(
+            nn.AdaptiveAvgPool2d(1), nn.Flatten(),
+            nn.Linear(self.in_channels, self.num_joints), nn.Sigmoid())
```
Tau-J, line 50: I suggest adding a `use_sigmoid: bool = False` in `__init__()` to determine wether to use `sigmoid` in `vis_head`, because Pytorch officially recommand to use `BCEWithLogitsLoss` instead of `Sigmoid + BCELoss`, which is "more numerically stable than using a plain Sigmoid followed by a BCELoss as, by combining the operations into one layer, we take advantage of the log-sum-exp trick for numerical stability." [details](https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html)

mmpose/datasets/transforms/formatting.py
```diff
@@ -99,7 +99,8 @@ class PackPoseInputs(BaseTransform):
         'keypoint_x_labels': 'keypoint_x_labels',
         'keypoint_y_labels': 'keypoint_y_labels',
         'keypoint_weights': 'keypoint_weights',
-        'instance_coords': 'instance_coords'
+        'instance_coords': 'instance_coords',
+        'keypoints_visible': 'keypoints_visible',
```
Tau-J, line 103: Could you please support `transformed_keypoints_visible` in https://github.com/open-mmlab/mmpose/blob/0796d77361201a862e9341d512b5445751df8832/mmpose/datasets/transforms/formatting.py#L156 
and https://github.com/open-mmlab/mmpose/blob/0796d77361201a862e9341d512b5445751df8832/mmpose/datasets/transforms/common_transforms.py#L935
because keypoint visibility will be used in `browse_dataset.py`

Tau-J (changes requested):
request changes

PR #2417, opened 2023-05-31
[Feature] Add visibility prediction head

Billccx:
## Motivation
Enable top-down methods to predict keypoints visibility

## Modification
1. add a new head `vis_head` in hybrid_heads      (mmpose/mmpose/models/heads/hybrid_heads/vis_head.py)
2. add a new TTA function `flip_visibility`     (mmpose/mmpose/models/utils/tta.py)
3. add a new key `transformed_keypoints_visible` in `label_mapping_table` (mmpose/mmpose/datasets/transforms/formatting.py)
4. support `keypoints_visible` in `GenerateTarget` (mmpose/mmpose/datasets/transforms/common_transforms.py)
5. <s>add a new loss function `BCEWithLogitsLoss` (mmpose/mmpose/models/losses/classification_loss.py)</s>
6.  refine `BCELoss` module: add parameter `with_logits` to decide whether to use BCEWithLogitsLoss  (mmpose/mmpose/models/losses/classification_loss.py)
7. add unittest for `vis_head`  (mmpose/tests/test_models/test_heads/test_hybrid_heads/test_vis_head.py)

## Use cases (Optional)

We can use `vis_head` to wrap a keypoints prediction head, getting the keypoints prediction results with keypoints visibility.

For example, in config file:

```
head=dict(
  type='VisPredictHead',
  loss=dict(type='BCELoss', use_target_weight=False, with_logits=True),
  use_sigmoid=False,
  pose_cfg=dict(
    type='HeatmapHead',
    in_channels=32,
    out_channels=17,
    deconv_out_channels=None,
    loss=dict(type='KeypointMSELoss', use_target_weight=True),
    decoder=codec
  )
)
```

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpose/models/heads/hybrid_heads/vis_head.py
+++ b/mmpose/models/heads/hybrid_heads/vis_head.py
@@ -168,6 +168,15 @@ class VisPredictHead(BaseHead):
 
         return self.integrate(batch_vis, batch_pose)
 
+    def vis_accuracy(self, vis_pred_outputs, vis_labels):
+        probabilities = torch.sigmoid(torch.flatten(vis_pred_outputs))
+        threshold = 0.5
+        predictions = (probabilities >= threshold).int()
+        labels = torch.flatten(vis_labels)
+        correct = torch.sum(predictions == labels).item()
+        accuracy = correct / len(labels)
+        return torch.tensor(accuracy)
+
     def loss(self,
              feats: Tuple[Tensor],
              batch_data_samples: OptSampleList,
@@ -198,6 +207,10 @@ class VisPredictHead(BaseHead):
                                     keypoint_weights)
 
         losses.update(loss_vis=loss_vis)
+
+        # calculate vis accuracy
+        acc_vis = self.vis_accuracy(vis_pred_outputs, vis_labels)
+        losses.update(acc_vis=acc_vis)
 
         # calculate keypoints losses
         loss_kpt = self.pose_head.loss(feats, batch_data_samples)
--- a/mmpose/models/heads/hybrid_heads/vis_head.py
+++ b/mmpose/models/heads/hybrid_heads/vis_head.py
@@ -30,7 +30,7 @@ class VisPredictHead(BaseHead):
     def __init__(self,
                  pose_cfg: ConfigType,
                  loss: ConfigType = dict(
-                     type='BCEWithLogitsLoss', use_target_weight=True),
+                     type='BCEWithLogitsLoss', use_target_weight=False),
                  use_sigmoid: bool = False,
                  init_cfg: OptConfigType = None):
 
@@ -52,6 +52,8 @@ class VisPredictHead(BaseHead):
 
         self.pose_head = MODELS.build(pose_cfg)
         self.pose_cfg = pose_cfg
+
+        self.use_sigmoid = use_sigmoid
 
         modules = [
             nn.AdaptiveAvgPool2d(1),
@@ -96,7 +98,10 @@ class VisPredictHead(BaseHead):
 
     def integrate(self, batch_vis: Tensor,
                   pose_preds: Union[Tuple, Predictions]) -> InstanceList:
-        """Add keypoints visibility prediction to pose prediction."""
+        """Add keypoints visibility prediction to pose prediction.
+
+        Overwrite the original keypoint_scores.
+        """
         if isinstance(pose_preds, tuple):
             pose_pred_instances, pose_pred_fields = pose_preds
         else:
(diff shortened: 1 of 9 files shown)


mmpose/models/heads/hybrid_heads/vis_head.py
```diff
@@ -0,0 +1,211 @@
...
+        else:
+            pose_pred_instances = pose_preds
+            pose_pred_fields = None
+
+        batch_vis_np = to_numpy(batch_vis, unzip=True)
+
+        assert len(pose_pred_instances) == len(batch_vis_np)
+        for index in range(len(pose_pred_instances)):
+            pose_pred_instances[index].keypoint_visibility = batch_vis_np[
+                index]
```
Ben-Louis, line 111: Why has a loop been utilized here? BTW, the visibility representation in 'pred_instances' is denoted by 'keypoint_scores'

Billccx: > Why has a loop been utilized here? BTW, the visibility representation in 'pred_instances' is denoted by 'keypoint_scores'

Thanks for your advise. I will rewrite this part soon, sorry for the bad code.

Ben-Louis: I'm sorry, my tone may have led to a misunderstanding. I was just unclear whether 'pose_pred_instances' is a tuple or InstanceData, so I wanted to ask if there is a possibility of not using a loop. Take it easy~

Billccx: > I'm sorry, my tone may have led to a misunderstanding. I was just unclear whether 'pose_pred_instances' is a tuple or InstanceData, so I wanted to ask if there is a possibility of not using a loop. Take it easy~

Got it! Sorry for the late reply. 

From my perspective, 'pose_pred_instances' is an 'InstanceList'? 

> https://github.com/open-mmlab/mmpose/blob/2c4a60e357c7f68f3388f0fc4ddcb404674b81ae/mmpose/models/heads/base_head.py#L78-L81

> https://github.com/open-mmlab/mmpose/blob/2c4a60e357c7f68f3388f0fc4ddcb404674b81ae/mmpose/models/heads/heatmap_heads/heatmap_head.py#L235-L241

Besides that, I have refined the code. If I still get something wrong, please let me know. Thanks.

mmpose/models/heads/hybrid_heads/vis_head.py
```diff
@@ -0,0 +1,228 @@
...
+            _batch_vis_flip = flip_visibility(
+                self.vis_forward(_feats_flip), flip_indices=flip_indices)
+            batch_vis = (_batch_vis + _batch_vis_flip) * 0.5
+        else:
+            batch_vis = self.vis_forward(feats)  # (B, K, D)
+
+        batch_vis.unsqueeze_(dim=1)  # (B, N, K, D)
+
+        if not self.use_sigmoid:
+            batch_vis = torch.sigmoid(batch_vis)
```
Ben-Louis, line 171: Why is `sigmoid` applied to `batch_vis` when `use_sigmoid` is `False`?

Billccx: I think the value of visibility prediction should be between 0 and 1? If `use_sigmoid` is `False`, there won't be sigmoid in `vis_forward`.

Ben-Louis: If the sigmoid function is always applied to the head outputs, what is the role of the `use_sigmoid` argument?

Tau-J: > Why is `sigmoid` applied to `batch_vis` when `use_sigmoid` is `False`?

@Ben-Louis When `use_sigmoid` is True, a `nn.Sigmoid()` will be in the `vis_head`, namely, the `batch_vis` is between 0~1. Otherwise, we need to do sigmoid manually when `use_sigmoid` is False.

Tau-J: Using `nn.Sigmoid()` and `BCELoss` separately is not officially recommended by Pytorch, a better way is to use `BCEWithLogitsLoss` and do sigmoid manually when a 0-1 score is needed. So I suggest him to set a `use_sigmoid` to keep the flexity.

Ben-Louis: @Tau-J, thanks for the clarification. This can indeed be puzzling, perhaps we could add some comments for clarity? Also, focusing on a single recommended practice could be beneficial. BTW, in MMDetection, this parameter is generally integrated in loss modules like `mmdet.CrossEntropyLoss`. We might consider refining the `BCELoss` module with this parameter, rather than introducing a completely new loss module.

Tau-J (approved):

PR #2531, opened 2023-07-10
compatible mmcv 1.7.1 and mmpose 0.29.0

liaochenchieh:
We want to use mmcv==1.7.1 and mmpose==0.29.0 at the same time to use the PCT algorithm (https://github.com/Gengzigang/PCT).
We've checked that these two versions can be used together when testing the PCT.

### DIFF
--- a/mmpose/__init__.py
+++ b/mmpose/__init__.py
@@ -17,7 +17,7 @@ def digit_version(version_str):
 
 
 mmcv_minimum_version = '1.3.8'
-mmcv_maximum_version = '1.7.0'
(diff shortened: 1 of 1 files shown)


Ben-Louis (approved):

Ben-Louis:
We suggest utilizing the `0.x` or `dev-0.x` branch instead of `dev-0.29` when using the 0.x version of mmpose. These branches have fixed the compatibility issue with mmcv-full 1.7.1.

As this issue has been resolved, we will temporarily close this PR. Feel free to reopen it if you have any additional updates.

PR #2584, opened 2023-07-30
[MMSIG] Support badcase analyze in test

Indigo6:
## Motivation

Analyze badcases in test, supported functions:

- [x] Set a sample as a badcase based on loss/accuracy
- [x] Show / store visualization of badcases
- [x] Store bbox and loss/accuracy info in img name and results json
- [x] Also supported draw line with str based color in `mmpose/visualization/opencv_backend_visualizer.py`

## Use cases (Optional)

Sample config file: `configs\body_2d_keypoint\topdown_heatmap\mpii\td-hm_mobilenetv2_8xb64-210e_mpii-256x256_badcase.py`

## Checklist

**Before PR**:

**After PR**:


configs/body_2d_keypoint/topdown_heatmap/mpii/td-hm_mobilenetv2_8xb64-210e_mpii-256x256_badcase.py
```diff
@@ -0,0 +1,124 @@
...
+auto_scale_lr = dict(base_batch_size=512)
+
+# hooks
+default_hooks = dict(checkpoint=dict(save_best='PCK', rule='greater'),
+                     badcase=dict(type="BadCaseAnalyzeHook", 
+                                #   metric_type="loss",
+                                  metric_type="accuracy",
+                                  show=True,
+                                  badcase_thr=100,
+                                  out_dir='badcase'))
```
Tau-J, line 36: How about moving `badcase` into `default_runtime.py` and disable by defaults? In this way, there is no need to add such a new config

configs/_base_/default_runtime.py
```diff
@@ -8,6 +8,11 @@
     checkpoint=dict(type='CheckpointHook', interval=10),
     sampler_seed=dict(type='DistSamplerSeedHook'),
     visualization=dict(type='PoseVisualizationHook', enable=False),
+    badcase=dict(type="BadCaseAnalyzeHook", 
+                 metric_type="loss",
+                 badcase_thr=100,
+                 show=True,
+                 out_dir='badcase')
```
Tau-J, line 15:
```suggestion
    badcase=dict(type="BadCaseAnalyzeHook", 
                 enable=False,
                 metric_type="loss",
                 badcase_thr=100,
                 show=True,
                 out_dir='badcase')
```

Tau-J: Should we set `enable=False` by defauts?

tools/test.py
```diff
@@ -88,6 +93,34 @@ def merge_args(cfg, args):
...
 
+    # -------------------- badcase analyze --------------------
+    if args.badcase:
+        assert 'badcase' in cfg.default_hooks, \
+            'BadcaseAnalyzeHook is not set in the ' \
+            '`default_hooks` field of config. Please set ' \
+            '`badcase=dict(type="BadcaseAnalyzeHook")`'
+        
+        cfg.default_hooks.badcase.enable = True
+        badcase_show = cfg.default_hooks.badcase.get('show', 'False')
```
Tau-J, line 104: Could we use `args.show` of `test.py` in badcase hook? I think the visualization of badcases could be a part of the visualization of prediction results

Indigo6: > Could we use `args.show` of `test.py` in badcase hook? I think the visualization of badcases could be a part of the visualization of prediction results, so it seems to make sense to share the same `show`

Will it cause showing two similar images? I think the visualization of badcase covers the visualization of prediction results. If use `args.show`, should we show badcase in default or add an user option choosing which one to show?

configs/_base_/default_runtime.py
Indigo6: Yes, I forgot this.

tools/test.py
Tau-J: Yeah, I guess it will result to display twice. Is it possible to merge the badcase analysis into `PoseVisualzationHook` to enhance it? Users can choose whether to enable the badcase visualization.

Indigo6: I suggest not, otherwise there will be too many args for `PoseVisualzationHook`, like `show_type(pred or badcase)`, 'store_badcase' .etc . I think we can add a `--show_type(pred or badcase)` option in argparser.

Tau-J: I think the behavior should be like:
1. `show=True`, `badcase=True`, show and store the badcase, no need to show the predictions
2. `show=False`, `badcase=True`, store the badcase
3.  `True`, `False`, show the prediction results
4. `False`, 'False', nothing

mmpose/engine/hooks/badcase_hook.py
```diff
@@ -0,0 +1,222 @@
...
+from mmengine.hooks import Hook
+from mmengine.runner import Runner
+from mmengine.visualization import Visualizer
+
+from mmpose.registry import HOOKS, MODELS, METRICS
+from mmpose.structures import PoseDataSample, merge_data_samples
+
+
+@HOOKS.register_module()
+class BadCaseAnalyzeHook(Hook):
```
Tau-J, line 22:
```suggestion
class BadCaseAnalysisHook(Hook):
```
It seems that `Analysis` is more reasonable?

tools/test.py
Indigo6: What if I just wanna analyze, store but not show badcase?

mmpose/engine/hooks/badcase_hook.py
Indigo6: OK. I didn't consider the name through.

tools/test.py
Tau-J: Yeah, that refers to `2. show=False, badcase=True`, and we need to set `badcase.show = args.show`

Indigo6: oh I missed that line …… This implementation logic is much greater

configs/body_2d_keypoint/topdown_heatmap/mpii/td-hm_mobilenetv2_8xb64-210e_mpii-256x256_badcase.py
Indigo6: Is the config of `badcase` in experiment config files able to overwrite that in default_runtime.py?

configs/_base_/default_runtime.py
Indigo6: Is it necessary since the default `enable` is `False` in `__init__` method?

mmpose/engine/hooks/badcase_hook.py
```diff
@@ -0,0 +1,226 @@
...
+                badcase_name = f'{img_name}_{bbox_info}_{metric_postfix}'
+
+                out_file = None
+                if self.out_dir is not None:
+                    out_file = f'{badcase_name}.{postfix}'
+                    out_file = os.path.join(self.out_dir, out_file)
+
+                # draw gt keypoints in blue color
+                self._visualizer.kpt_color = 'blue'
+                self._visualizer.link_color = 'blue'
```
Tau-J, line 181:
```suggestion
                # draw gt keypoints in green color
                self._visualizer.kpt_color = 'green'
                self._visualizer.link_color = 'green'
```

mmpose/engine/hooks/badcase_hook.py
```diff
@@ -0,0 +1,226 @@
...
+                from the dataloader.
+            data_samples (Sequence[dict]): A batch of outputs from
+                the model.
+        Return:
+            is_badcase (bool): whether the sample is a badcase or not
+            metric_value (float)
+        """
+        if self.metric_type == 'loss':
+            gts = data_sample.gt_instances.keypoints
+            preds = data_sample.pred_instances.keypoints
```
Tau-J, line 118:
```suggestion
            gts = data_sample.gt_instances.keypoints
            preds = data_sample.pred_instances.keypoints
            weights = data_sample.gt_instances.keypoints_visible
```

mmpose/engine/hooks/badcase_hook.py
```diff
@@ -0,0 +1,226 @@
...
+            with torch.no_grad():
+                metric_value = self.metric(
+                    torch.tensor(preds), torch.tensor(gts)).item()
```
Tau-J, line 121:
```suggestion
                    torch.from_numpy(preds), torch.from_numpy(gts), torch.from_numpy(weights)).item()
```

mmpose/visualization/opencv_backend_visualizer.py
```diff
@@ -358,7 +358,8 @@ def draw_lines(self,
                 **kwargs)
 
         elif self.backend == 'opencv':
-
+            if isinstance(colors, str):
+                colors = mmcv.color_val(colors)
```
Tau-J, line 362:
```suggestion
                colors = mmcv.color_val(colors)[::-1]
```

Tau-J: I notice that `mmcv.color_val()` only support `bgr` mode, yet mmpose use `rgb` to visualize. So could you please also modify all `mmcv.color_val(colors)` in `opencv_backend_visualizer`?

mmpose/engine/hooks/badcase_hook.py
Indigo6: Um …… I suggest using `blue` for easy distinction

mmpose/engine/hooks/badcase_hook.py
Indigo6: Ok! then use weights in `loss` metric cal, right?

mmpose/engine/hooks/badcase_hook.py
Indigo6: Ok!

mmpose/visualization/opencv_backend_visualizer.py
Indigo6: Ok, I see.

Tau-J (approved):

PR #2736, opened 2023-10-07
[Feature] Support H3WB dataset

xiexinch:
## Motivation

Support 3d wholebody dataset H3WB.

[Official webpage](https://github.com/wholebody3d/wholebody3d)

## Modification

- Modify mmpose/apis/inference_3d.py to support inference api

## Checklist

**Before PR**:

**After PR**:


mmpose/datasets/datasets/wholebody3d/h3wb_dataset.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+
+    METAINFO: dict = dict(from_file='configs/_base_/datasets/h3wb.py')
+
+    def __init__(self,
+                 ann_file: str,
+                 data_root: str,
+                 data_prefix: dict,
+                 joint_2d_src: str,
+                 normalize_with_dataset_stats: bool = False,
```
Ben-Louis, line 99: Please add description for these arguments in docstring

mmpose/datasets/transforms/pose3d_transforms.py
```diff
@@ -44,11 +44,11 @@ class RandomFlipAroundRoot(BaseTransform):
     """
 
     def __init__(self,
-                 keypoints_flip_cfg,
-                 target_flip_cfg,
-                 flip_prob=0.5,
-                 flip_camera=False,
-                 flip_label=False):
+                 keypoints_flip_cfg: dict,
+                 target_flip_cfg: dict,
```
Ben-Louis, line 48: Do you think it would be better to pass arguments directly instead of using a dict?

xiexinch: > Do you think it would be better to pass arguments directly instead of using a dict?

I think it's better to create a new PR if we decide to list all parameters at the constructor, it will affect all MotionBERT and VideoPose configs.

Ben-Louis: > I think it's better to create a new PR if we decide to list all parameters at the constructor, it will affect all MotionBERT and VideoPose configs.

All right. Just keep it this time

mmpose/datasets/datasets/wholebody3d/h3wb_dataset.py
```diff
@@ -0,0 +1,248 @@
...
+            image. Default: 1000.
+    """
+
+    METAINFO: dict = dict(from_file='configs/_base_/datasets/h3wb.py')
+
+    def __init__(self, ann_file: str, data_root: str, joint_2d_src: str,
+                 data_prefix: dict, **kwargs):
+        self.ann_file = ann_file
+        self.data_root = data_root
+        self.data_prefix = data_prefix
```
Ben-Louis, line 97: Is it necessary to assign these arguments manually? They will be assigned in `BaseDataset.__init__`

Ben-Louis (approved):

PR #2756, opened 2023-10-14
[Fix] Let RTMPose onnxruntime example support 17 keypoints model

RFYoung:
## Motivation

In the onnxruntime example for RTMPose, I will encounter an error when using 17 keypoints models. I have made a tiny fix for this problem. 

## Modification

When drawing the lines, the program would ignore the point index from the `skeleton` that is out of the range of model output.

## Checklist

**Before PR**:

**After PR**:


RFYoung:
Here is an output using RTMPose tiny (body 8, onnx) with 17 keypoints.
output

Tau-J (approved):

PR #2961, opened 2024-02-21
[Feature] add wflw2coco

Yanyirong:
## Motivation

Add wflw2coco script, which transform wflw to coco dataset type.

Although it is not same to the originally processed wflw annotations, but the key values are the same and the performance is correct.

### DIFF
--- a/docs/en/dataset_zoo/2d_face_keypoint.md
+++ b/docs/en/dataset_zoo/2d_face_keypoint.md
@@ -152,7 +152,7 @@ mmpose
             ...
 ```
 
-We also supply a script which can convert the raw WFLW dataset to COCO style. The data processed by the script is not completely consistent with the data downloaded from the URL, but it does not affect training and testing.
+We also supply a script which can convert the raw WFLW annotations to COCO style. The script's output is not completely consistent with the annotations downloaded from the URL, but it does not affect training and testing.
 
 ## AFLW Dataset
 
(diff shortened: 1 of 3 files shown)


docs/en/dataset_zoo/2d_face_keypoint.md
```diff
@@ -152,6 +152,8 @@ mmpose
             ...
 ```
 
+We also supply a script which can convert the raw WFLW dataset to COCO style. The data processed by the script is not completely consistent with the data downloaded from the URL, but it does not affect training and testing.
```
Ben-Louis, line 155:
```suggestion
We also supply a script which can convert the raw WFLW annotations to COCO style. The script's output is not completely consistent with the annotations downloaded from the URL, but it does not affect training and testing.
```

docs/zh_cn/dataset_zoo/2d_face_keypoint.md
```diff
@@ -152,6 +152,8 @@ mmpose
             ...
 ```
 
+We also supply a script which can convert the raw WFLW dataset to COCO style. The data processed by the script is not completely consistent with the data downloaded from the URL, but it does not affect training and testing.
```
Ben-Louis, line 155:
```suggestion
We also supply a script which can convert the raw WFLW annotations to COCO style. The script's output is not completely consistent with the annotations downloaded from the URL, but it does not affect training and testing.
```

Ben-Louis (approved):

Issue #3199, opened 2025-03-18
[Bug] mmpose/mmpose/visualization/local_visualizer_3d.py


lychenpan:
### Prerequisite

- [x] I have searched [Issues](https://github.com/open-mmlab/mmpose/issues) and [Discussions](https://github.com/open-mmlab/mmpose/discussions) but cannot get the expected help.
- [x] The bug has not been fixed in the latest version(https://github.com/open-mmlab/mmpose).

### Environment

![Image](https://github.com/user-attachments/assets/939613bc-c784-433f-9f0e-5967140dfedf)
这个代码没有问题么  

### Reproduces the problem - code sample

no need to reproduce

### Reproduces the problem - command or script

no need to reproduce

### Reproduces the problem - error message

no need to reproduce

PR #3268, opened 2026-06-29
[Fix] 3-arg np.logical_and drops NaN-keypoint filter in 3D visualizer (#3199)

yushuosun:
## Motivation

The 3D pose visualizer's keypoint visibility mask is computed incorrectly, so NaN keypoints are not filtered out as intended (#3199).

## Root cause

In `mmpose/visualization/local_visualizer_3d.py`, `_draw_3d_instances_kpts` builds the mask with a 3-argument `np.logical_and`:

```python
valid = np.logical_and(score >= kpt_thr, score_2d >= kpt_thr,
                       np.any(~np.isnan(kpts), axis=-1))
```

`np.logical_and` is a binary ufunc — its third positional argument is the `out` buffer, not a third condition. So the intended NaN-filter term `np.any(~np.isnan(kpts), axis=-1)` is written into `out` and the result merely aliases it; the `score`/`score_2d` thresholds and the NaN filter are not all AND-ed together.

## Modifications

`mmpose/visualization/local_visualizer_3d.py`: combine the three conditions with `np.logical_and.reduce([...])` so all of them are applied. (The two-argument `np.logical_and` uses in `local_visualizer.py` are correct and untouched.)

## Duplicate-check

- Track A — `gh api 'repos/open-mmlab/mmpose/pulls?state=open&per_page=100' --paginate` → no open PR references #3199 or touches `local_visualizer_3d.py`.
- Track B — issue #3199 has no in-progress claim.


yushuosun:
Closing because the repository has been inactive for about a year, the CLA is still pending, and CI is failing without maintainer engagement. The patch remains on the branch if development resumes.

